Issue #22, opened 2019-09-10
Workflow creation screen update


Zejnilovic:
Here is the screen that was discussed between @jozefbakus and me.
![image](https://user-images.githubusercontent.com/17467768/64610752-38233300-d3d0-11e9-8f6d-af76d04424a9.png)

It is mostly moving around screens but one thing that changes a lot is the Job name is now at the top of the screen not bellow Job type and it should not be deleted if a person switches job type.

This allows for multiple job definitions. Same way sensors could be handled if we decide to have more then one per workflow

PR #33, opened 2019-10-03
Workflow with multiple jobs

jozefbakus:
Workflow with multiple jobs
Issue: https://github.com/AbsaOSS/hyperdrive-trigger/issues/22


src/main/resources/ui/controller/job/SparkController.js
```diff
@@ -0,0 +1,17 @@
...
+
+    onDeleteAppArgument(oEv) {
+        UiListMethods.deleteListItem("/newJob/job/jobParameters/maps/appArguments", this._model, oEv)
+    }
+
+    onAddAppArgument() {
+        UiListMethods.addListItem("/newJob/job/jobParameters/maps/appArguments", this._model, "");
+    }
+
+}
```
Zejnilovic, line 17: I see the `.editorconfig` might be needed

PR #35, opened 2019-10-09
Added copyright to files

jozefbakus:
Added copyright to files.


kevinwallimann:
I would add the copyright also here:
- `src/main/resources/application.properties`
- `src/main/resources/logback.xml`

In the main `pom.xml`, the copyright is only 2019 instead of 2018-2019.

Issue #41, opened 2019-11-18
LDAP authentication 


jozefbakus:
Use LDAP to authenticate.

PR #42, opened 2019-11-18
Feature/ldap authentication

jozefbakus:
Use LDAP to authenticate. (Issue: #41)
Merge two configs to one.
Add dialog to confirm workflow run via UI.
Refresh runs bug fix.


src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/utilities/Configs.scala
```diff
@@ -29,13 +29,14 @@ private object Configs {
...
-      } yield (key, value)).toMap
+      def getKeys(path: String): Seq[String] = {
+        Try(Configs.conf.getObject(path).keySet().asScala).getOrElse(Set.empty[String]) match {
+          case keys if keys.nonEmpty => keys.flatMap(k => getKeys(s"$path.$k")).toSeq
+          case keys if keys.isEmpty => Seq(path)
+        }
+      }
+      val keys = getKeys(propertyName)
+      keys.map(k => (k.stripPrefix(s"$propertyName."), conf.getString(k))).toMap
     }.getOrElse(Map.empty[String, String])
```
kevinwallimann, line 40: A test would be good for this non-trivial logic. Also, is it not possible to get all keys and then just filter them? It seems over-complicated to have recursion here. (Not your fault I guess. Seems like `typesafe.config` is not the simplest library out there)

pom.xml
```diff
@@ -119,7 +120,6 @@
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-tomcat</artifactId>
             <version>${spring-boot-starter-tomcat.version}</version>
-            <scope>provided</scope>
```
kevinwallimann, line 122: I don't understand this change. Why does the scope need to be `compile` now?

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/auth/LdapAuthentication.scala
```diff
@@ -0,0 +1,62 @@
...
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.security.config.annotation.authentication.builders.AuthenticationManagerBuilder
+import org.springframework.security.ldap.authentication.ad.ActiveDirectoryLdapAuthenticationProvider
+import org.springframework.stereotype.Component
+
+@Component
+class LdapAuthentication extends HyperdriverAuthentication {
+
+  @Value("${auth.ad.domain:}")
+  val adDomain: String = ""
```
kevinwallimann: Is the ` = ""` needed? You already specify a default value with the `:` in the `@Value`

src/main/resources/application.properties
```diff
@@ -5,4 +5,48 @@ logging.file=${logging.path}/app.log
 
 management.endpoints.web.base-path=/admin
 management.endpoints.jmx.exposure.exclude=*
-management.endpoints.web.exposure.include=*
\ No newline at end of file
+management.endpoints.web.exposure.include=*
```
Zejnilovic: 1. missing a new line
2. shouldn't you provide only template and not the actual properties?
3. Why is everything in auth commented out?
4. Provide more comments for the users as to what is what

src/main/resources/ui/controller/Workflows.controller.js
```diff
@@ -77,6 +77,18 @@ sap.ui.define([
 			});
 		},
 
+		createConfirmRunWorkflowDialog: function (id) {
+			MessageBox.confirm("Are you sure you want to run workflow ?", {
```
Zejnilovic:
```suggestion
			MessageBox.confirm("Are you sure you want to run workflow?", {
```

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/auth/LdapAuthentication.scala
Zejnilovic, line 27: 1. In spring doc's there is always a newline between 2 props.
2. If the prop is missing, shouldn't there be an automatic error about it? Meaning you wouldn't have to validate.

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/utilities/Configs.scala
```diff
@@ -29,13 +29,14 @@ private object Configs {
...
   def getMapFromConf(propertyName: String): Map[String, String] = {
     Try {
-      val list = Configs.conf.getObjectList(propertyName).asScala
-      (for {
-        item: ConfigObject <- list
-        entry <- item.entrySet().asScala
-        key = entry.getKey
-        value = entry.getValue.unwrapped().toString
-      } yield (key, value)).toMap
+      def getKeys(path: String): Seq[String] = {
```
Zejnilovic, line 32: Have you seen this [API DOC](https://lightbend.github.io/config/latest/api/com/typesafe/config/ConfigObject.html#unwrapped--)

pom.xml
jozefbakus: It was your suggestion here https://github.com/AbsaOSS/hyperdrive/pull/10#discussion_r318432678
This dep. will be removed soon.

src/main/resources/application.properties
jozefbakus: 1. Where are you missing a new line ? ... Removed/Fixed
2. True, management props are just for us (debugging) ... Removed/Fixed
3. Ok ... inmemory is uncommented now. ... Fixed
4. OK ... Fixed

src/main/resources/ui/controller/Workflows.controller.js
jozefbakus: Fixed.

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/auth/LdapAuthentication.scala
jozefbakus: 1. Fixed. 
2. Yes, that is true, but if I leave it in spring it shows `Caused by: java.lang.IllegalArgumentException: Url cannot be empty` which is not that good as `Caused by: java.lang.IllegalArgumentException: auth.ad.domain has to be configured in order to use ldap authentication`.

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/utilities/Configs.scala
jozefbakus: You are right. I have created issue #43 for that and it has to be properly handled.

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/utilities/Configs.scala
jozefbakus: You are right. The same answer as for Sasa:  I have created issue #43 for that and it has to be properly handled.

src/main/resources/application.properties
Zejnilovic: 1. End of the file

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/auth/LdapAuthentication.scala
Zejnilovic: 2. Are you saying that `validateParams` will be called before the springs natural error?

jozefbakus: In general no, but in this case, a default value is provided.

src/main/resources/application.properties
jozefbakus: 1. Done.

Zejnilovic (approved):
Thank you

jozefbakus:
> Thank you

No, no, no ... THANK YOU!

kevinwallimann (approved):

Issue #46, opened 2019-11-25
Angular initial skeleton


kevinwallimann:
Purpose of this PR: 
- Starting point for further Angular development

What has been done:
- Authentication (can still be simplified, but requires changes in server config that would break the current ui)
- Global error handling
- Notifications (aka Alerts)
- Unit-Tests
- Dev-Server Setup (Webpack)

No component library has been chosen yet, therefore the look-and-feel is extremely poor.

**Compatibility of PR with old UI**
This PR can be merged without affecting the existing ui, but the following configuration changes must be undone for the old ui. That is
- `spring.resources.static-locations` back to `classpath:/ui/`
- `rest.api.prefix` set to ` ` (empty string)

PR #47, opened 2019-11-26
Feature/ui angular

kevinwallimann:
See #46


kevinwallimann:
Notes maven integration (27360b9)

- Moved source js files out of resources folder to reduce jar size. Now at /src/main/ui-angular
- Angular compilation target points not to dist anymore, but target/classes/ui-angular
- `mvn clean install` downloads npm and node (to local folder, no conflict with existing installations) builds and tests the angular project
- `-Dskip.npm` and `-Dskip.installnodenpm` can be used to skip frontend build.
- jar has to be run with `rest.api.prefix=/api` and `spring.resources.static-locations=classpath:/ui-angular/`

See /src/main/ui-angular/README.md for more instructions.

Issue #52, opened 2019-12-06
Use Dependency Injection instead of manual object creation


kevinwallimann:
Currently, all repositories are created by manual object creation in `HyperDriverManager.initialize`. This should be changed in favor of Spring Dependency Injection.

kevinwallimann:
Closed because PR merged

Issue #51, opened 2019-12-04
@Inject vs @Autowired


jozefbakus:
@Inject vs @Autowired

kevinwallimann:
https://javarevisited.blogspot.com/2017/04/difference-between-autowired-and-inject-annotation-in-spring-framework.html

- @Inject is specified by JSR-330, @Autowired isn't
- Since Spring 3.0, both are supported
- @Autowired seems to be more popular, but that may be just because many projects started before Spring 3.0
=> I vote for @Inject

kevinwallimann:
Closed because PR merged

PR #53, opened 2019-12-06
Use Spring DI with @Inject keyword instead of manual object creation …

kevinwallimann:
…(#52, #51)

This solves #52  and #51


jozefbakus (approved):
I love it !

PR #56, opened 2019-12-10
Feature/time trigger 2

kevinwallimann:
See #48 . This is a draft.

Missing:
- Separate thread pool for quartz jobs
- Integration test for TimeSensor

PR #57, opened 2019-12-10
Feature/time trigger ui

jozefbakus:
Time trigger fragment with simple validation (will not let you save workflow with incorrect cron expression).
Related to backend implementation: #56

### DIFF
--- a/src/main/resources/ui/controller/UpsertWorkflow.controller.js
+++ b/src/main/resources/ui/controller/UpsertWorkflow.controller.js
@@ -104,14 +104,27 @@ sap.ui.define([
 		},
 
 		onSaveWorkflow: function () {
-			let isEdit = this._model.getProperty("/isEdit");
-			let workflow = this.getWorkflowToSave();
-			if(isEdit) {
-				WorkflowRepository.updateWorkflow(workflow);
-			} else {
-				WorkflowRepository.createWorkflow(workflow);
+			if(this._isWorkflowValid(this._model.getProperty("/workflow"))){
+				let isEdit = this._model.getProperty("/isEdit");
+				let workflow = this.getWorkflowToSave();
+				if(isEdit) {
+					WorkflowRepository.updateWorkflow(workflow);
+				} else {
+					WorkflowRepository.createWorkflow(workflow);
+				}
+				this.myNavBack();
 			}
-			this.myNavBack();
+		},
+
+		_isWorkflowValid: function (workflow) {
+			try {
+				cronstrue.toString(workflow.sensor.properties.settings.variables.cronExpression);
+			}
+			catch(err) {
+				new sap.m.MessageToast.show("Invalid CRON expression", {animationDuration: 5000});
+				return false;
+			}
+			return true;
 		},
 
 		getWorkflowToSave: function () {
--- a/src/main/resources/ui/controller/UpsertWorkflow.controller.js
+++ b/src/main/resources/ui/controller/UpsertWorkflow.controller.js
@@ -20,7 +20,6 @@ sap.ui.define([
 ], function (BaseController, Fragment, Filter) {
 	"use strict";
 	return BaseController.extend("hyperdriver.controller.UpsertWorkflow", {
-
 		onInit: function () {
 			this.getRouter().attachRouteMatched(this.onViewDisplay, this);
 			this.getView().setModel(new sap.ui.model.json.JSONModel());
@@ -104,7 +103,7 @@ sap.ui.define([
 		},
 
 		onSaveWorkflow: function () {
-			if(this._isWorkflowValid(this._model.getProperty("/workflow"))){
+			if(this._isWorkflowValid()){
 				let isEdit = this._model.getProperty("/isEdit");
 				let workflow = this.getWorkflowToSave();
 				if(isEdit) {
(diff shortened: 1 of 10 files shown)


src/main/resources/ui/view/sensor/time.fragment.xml
```diff
@@ -0,0 +1,25 @@
...
+  -->
+
+<core:FragmentDefinition xmlns="sap.m" xmlns:core="sap.ui.core">
+    <Label text="Will run"/>
+    <Text text="{parts: ['/workflow/sensor/properties/settings/variables/cronExpression'],
+                 formatter: 'Formatters.cronFormatter'}" />
+    <Label text="Cron Expression"/>
+    <Input type="Text"
+           value="{/workflow/sensor/properties/settings/variables/cronExpression}"
+           placeholder="* * * * * *"
```
kevinwallimann: Isn't this the invalid expression?

src/main/resources/ui/controller/UpsertWorkflow.controller.js
```diff
@@ -104,14 +104,27 @@ sap.ui.define([
...
+					WorkflowRepository.createWorkflow(workflow);
+				}
+				this.myNavBack();
 			}
-			this.myNavBack();
+		},
+
+		_isWorkflowValid: function (workflow) {
+			try {
+				cronstrue.toString(workflow.sensor.properties.settings.variables.cronExpression);
```
kevinwallimann: You could add a custom validation here for that specific case that cronstrue checks wrongly (one of day of month and day of week must be `?`)

kevinwallimann: In fact, please do

kevinwallimann (changes requested):

src/main/resources/ui/controller/UpsertWorkflow.controller.js
```diff
@@ -104,14 +104,27 @@ sap.ui.define([
...
-				WorkflowRepository.createWorkflow(workflow);
+			if(this._isWorkflowValid(this._model.getProperty("/workflow"))){
+				let isEdit = this._model.getProperty("/isEdit");
+				let workflow = this.getWorkflowToSave();
+				if(isEdit) {
+					WorkflowRepository.updateWorkflow(workflow);
+				} else {
+					WorkflowRepository.createWorkflow(workflow);
+				}
+				this.myNavBack();
```
Zejnilovic: The tabbing seems extensive. How about spaces and a max of 4?

src/main/resources/ui/controller/UpsertWorkflow.controller.js
```diff
@@ -186,7 +199,8 @@ sap.ui.define([
 
 		sensorTypes: [
 			{name: "Absa-Kafka", fragment: "absaKafka", controller: "AbsaKafkaController"},
-			{name: "Kafka", fragment: "kafka", controller: "KafkaController"}
+			{name: "Kafka", fragment: "kafka", controller: "KafkaController"},
+			{name: "Time", fragment: "time", controller: "TimeController"}
```
Zejnilovic, line 201: This should be slowly moved to configuration

src/main/resources/ui/controller/UpsertWorkflow.controller.js
Zejnilovic: If you do it as in Enceladus and let user only select what is possible, then you can skip the validations. 

BTW Why don't you use the output of the `toString` function to show to the users, so they know what time they have set up. This is one of the really convenient features for the user

src/main/resources/ui/controller/UpsertWorkflow.controller.js
jozefbakus: Will be fixed here: #60

src/main/resources/ui/controller/UpsertWorkflow.controller.js
jozefbakus: Will be fixed here: #60

src/main/resources/ui/controller/UpsertWorkflow.controller.js
jozefbakus: Kevin - yes, fixed
Sasa - 1. Yes in Angular UI. 2. Yes, we do.

src/main/resources/ui/view/sensor/time.fragment.xml
jozefbakus: Yes, fixed

src/main/resources/ui/controller/sensor/TimeController.js
```diff
@@ -29,8 +29,15 @@ class TimeController {
         let cronExpressionPath = "/workflow/sensor/properties/settings/variables/cronExpression";
         let cronExpression = this._model.getProperty(cronExpressionPath);
         if(!cronExpression) {
-            this._model.setProperty(cronExpressionPath, "0 0 * * * *")
+            this._model.setProperty(cronExpressionPath, "0 0 0 ? * * *")
```
kevinwallimann, line 32: Did you change this intentionally from a 6-character to a 7-character cron expression?

kevinwallimann (approved):

src/main/resources/ui/controller/sensor/TimeController.js
jozefbakus: Yes, 7th character is optional, but the expression is still valid.

Issue #58, opened 2019-12-12
DB connection pool - multiple pools are created


jozefbakus:
DB connection pool - multiple pools are created.
For each repository implementation, a new thread pool is created. 
All connections to DB should be managed by one thread pool.

PR #59, opened 2019-12-12
DB thread pool fix.

jozefbakus:
Only one thread pool is used for DB connections.
#58


kevinwallimann (approved):
Awesome that you noticed it. That's a pretty subtle bug

PR #61, opened 2019-12-18
Provided prop file fix

jozefbakus:
We can provide properties file which is not compiled within jar/war, using -Dspring.config.location=/aa/bbb.properties

### DIFF
--- a/src/main/resources/logback.xml
+++ b/src/main/resources/logback.xml
@@ -1,13 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration>
 
-    <property resource="application.properties" />
-
     <appender name="FILE-ROLLING" class="ch.qos.logback.core.rolling.RollingFileAppender">
-        <file>${logging.file}</file>
+        <file>logs/app.log</file>
 
         <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
-            <fileNamePattern>${logging.path}/app.%d{yyyy-MM-dd}.%i.log.gz</fileNamePattern>
+            <fileNamePattern>logs/app.%d{yyyy-MM-dd}.%i.log.gz</fileNamePattern>
             <!-- each archived file, size max 15MB -->
             <maxFileSize>15MB</maxFileSize>
             <!-- total size of all archive files, if total size > 20GB, it will delete old archived file -->
--- a/src/main/resources/logback.xml
+++ b/src/main/resources/logback.xml
@@ -7,7 +7,7 @@
         <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
             <fileNamePattern>logs/app.%d{yyyy-MM-dd}.%i.log.gz</fileNamePattern>
             <!-- each archived file, size max 15MB -->
-            <maxFileSize>15MB</maxFileSize>
+            <maxFileSize>512MB</maxFileSize>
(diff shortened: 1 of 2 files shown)


src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/utilities/Configs.scala
```diff
@@ -15,17 +15,22 @@
...
 import scala.collection.JavaConverters._
 import scala.util.Try
 
 private object Configs {
-  val conf: Config = ConfigFactory.load()
+  private val configPath: Option[String] = Option(System.getProperty("spring.config.location")).filter(_.trim.nonEmpty)
+  val conf: Config = configPath match {
+    case Some(cp) => ConfigFactory.parseFile(new File(cp))
+    case None => ConfigFactory.load()
+  }
```
kevinwallimann, line 33: I think it would be better if the user defined configuration is combined with the default configuration.
Then, for each environment, only the properties that are different for that environment have to be specified

src/main/resources/logback.xml
```diff
@@ -1,13 +1,11 @@
...
-
     <appender name="FILE-ROLLING" class="ch.qos.logback.core.rolling.RollingFileAppender">
-        <file>${logging.file}</file>
+        <file>logs/app.log</file>
 
         <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
-            <fileNamePattern>${logging.path}/app.%d{yyyy-MM-dd}.%i.log.gz</fileNamePattern>
+            <fileNamePattern>logs/app.%d{yyyy-MM-dd}.%i.log.gz</fileNamePattern>
             <!-- each archived file, size max 15MB -->
             <maxFileSize>15MB</maxFileSize>
```
kevinwallimann: Unrelated, but isn't 15 MB a bit small? This results in many small files which can be annoying if you want to analyze the log file

src/main/resources/logback.xml
```diff
@@ -1,13 +1,11 @@
...
 
-    <property resource="application.properties" />
-
     <appender name="FILE-ROLLING" class="ch.qos.logback.core.rolling.RollingFileAppender">
-        <file>${logging.file}</file>
+        <file>logs/app.log</file>
 
         <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
-            <fileNamePattern>${logging.path}/app.%d{yyyy-MM-dd}.%i.log.gz</fileNamePattern>
+            <fileNamePattern>logs/app.%d{yyyy-MM-dd}.%i.log.gz</fileNamePattern>
```
kevinwallimann, line 8: Why hard-code the logging path?

jozefbakus: As discussed, this one is fine.

src/main/resources/logback.xml
jozefbakus: Fixed.

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/utilities/Configs.scala
jozefbakus: This is how spring works. We can override spring boot process, but it would be a hacky way. 
I prefer this one.

kevinwallimann (approved):

src/main/resources/logback.xml
kevinwallimann: For the record, if the logging path should be customized, a custom logback configuration should be provided

Issue #74, opened 2020-01-28
Containerization: Dev setup with DB


kevinwallimann:
- Add docker image for hyperdrive-trigger.
- Additionally, provide support for a Postgres DB. This can be done e.g. with docker-compose or bundled within the image.

The scope of this issue is only the application and the database. The docker composition should for now only be used for development. For a full-fledged production use-case, the hyperdrive jar needs to be added and support for Spark and Hadoop. Furthermore, logs should be stored in a volume and the volumes must be backed up. Also, credentials must be handled in a secure way.

PR #75, opened 2020-01-28
Add docker scripts for development purposes (app + db) (#74)

kevinwallimann:


docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This docker-compose script is only for development purposes! Passwords should never be stored in plaintext!
+version: "3"
+
+services:
```
Zejnilovic, line 17: Just a quick one. Don't you have to create a network to join services likes this? 
What I mean at docker [docs](https://docs.docker.com/compose/compose-file/#networks)

kevinwallimann: Yeah no. A default network is created by docker-compose. https://docs.docker.com/compose/networking/

jozefbakus (commented):
This is perfect!

PR #78, opened 2020-01-31
Make scalatest plugin respect mvn flag -DskipTests

kevinwallimann:


jozefbakus (approved):
👍Working as expected.

PR #80, opened 2020-02-10
Spark - additional files

jozefbakus:
Additional files for spark executor.
Start the scheduler at the start of the application.


kevinwallimann (approved):
Changes are fine for me. 
For the concrete use case, as I said, I would provide a method to import a properties file and then the app would transform it, such that the properties are saved in the same way as the jobs that use CommandLineIngestor. I think we should only provide one interface to configure hyperdrive, otherwise we might end up with inconsistencies. But that would be a change for the redesigned app.

PR #82, opened 2020-02-26
LogIn screen + Auth lifecycle using ngrx

jozefbakus:


ui/src/app/app.component.ts
```diff
@@ -1,10 +1,35 @@
...
+export class AppComponent implements OnInit, OnDestroy {
+
+  authStateSubscription: Subscription;
+  isAuthenticated: boolean;
+  username: string;
+
+  constructor(private store: Store<fromApp.AppState>) {}
+
+  ngOnInit(): void {
+    this.authStateSubscription = this.store.select('auth').subscribe((state) => {
```
kevinwallimann: I would extract the constant `'auth'` to a variable. Maybe put it in `app.reducers.ts` to have everything in the same place?

ui/src/app/services/interceptors/csrf.interceptor.ts
```diff
@@ -0,0 +1,18 @@
+import {Injectable} from '@angular/core';
```
kevinwallimann, line 16: Maybe consider using the `HttpClientXsrfModule` from Angular then we don't have to care about the csrf-token ourselves. Some changes are required on the backend. I think, this would best be done in another issue. I'd volunteer to work on it.

ui/src/app/services/auth/auth.service.ts
```diff
@@ -0,0 +1,53 @@
...
+      map(_ => _.username)
+    );
+  }
+
+  login(username: string, password: string): Observable<string> {
+    const body = new FormData();
+    body.append('username', username);
+    body.append('password', password);
+    return this.httpClient.post(
+      '/api/login',
```
kevinwallimann: I think these paths should not be hard-coded

ui/src/app/stores/auth/auth.effects.ts
```diff
@@ -0,0 +1,79 @@
...
+      localStorage.setItem('csrf-token', action.payload.token);
+      localStorage.setItem('username', action.payload.username);
+      this.router.navigateByUrl('/');
+    })
+  );
+
+  @Effect({dispatch: true})
+  logOut: Observable<any> = this.actions.pipe(
+    ofType(AuthActions.LOGOUT),
+    switchMap(_ => this.authService.logout()),
```
kevinwallimann, line 67: Here we will need some error handling, if `authService.logout` returns an error. I guess you omitted it because we don't have general error handling yet.

ui/src/app/app.component.ts
kevinwallimann: Adressed in #85

ui/src/app/services/auth/auth.service.ts
kevinwallimann: Adressed in #85

ui/src/app/services/interceptors/csrf.interceptor.ts
kevinwallimann: Not considering `HttpClientXsrfModule` and cookie-to-header method because cookie can't be read from different port in live-serving mode. The current localStorage-to-header method is equally secure.

kevinwallimann (approved):

Issue #96, opened 2020-03-12
Workflows - main screen - side navigation


jozefbakus:
Workflows - main screen - side navigation.
- plus link to main workflows screen. (All workflows or something similar)
- Using backend and ngrx.

PR #102, opened 2020-03-17
Workflows - main screen - side navigation #96

jozefbakus:
implements: #96


kevinwallimann:
I would name the branches like this: feature/{issue number}-{some description}, so e.g. feature/96-workflows-navigation.

kevinwallimann:
I think the sidenav should be resizable. For really long workflow names, there will be multiple word wraps, while there would be enough space on the right side, especially on a big screen

ui/src/app/components/workflows/workflows.component.ts
```diff
@@ -13,18 +13,41 @@
...
+  projects: ProjectModel[] = [];
+  workflows: WorkflowModel[] = [];
 
-  ngOnInit(): void {
+  constructor(private store: Store<AppState>) {
+    this.store.dispatch(new InitializeWorkflows());
+  }
+
+  ngAfterViewInit(): void {
+    this.workflowsSubscription = this.store.select(selectWorkflowState).pipe(skip(1)).subscribe((state) => {
```
kevinwallimann: What is the `skip(1)` for?

ui/src/app/stores/workflows/workflows.effects.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {TestBed} from '@angular/core/testing';
+// import {RunsEffects} from "./work.effects";
```
kevinwallimann: Some unused imports

kevinwallimann (commented):
Please see my comments. Code is really good, I've got almost nothing to say

ui/src/app/components/workflows/workflows.component.ts
kevinwallimann: Nevermind. (Removed in #103)

ui/src/app/stores/workflows/workflows.effects.spec.ts
jozefbakus: Done

jozefbakus:
> I would name the branches like this: feature/{issue number}-{some description}, so e.g. feature/96-workflows-navigation.

With new branches, I will follow this.

jozefbakus:
> I think the sidenav should be resizable. For really long workflow names, there will be multiple word wraps, while there would be enough space on the right side, especially on a big screen

This was partially fixed here https://github.com/AbsaOSS/hyperdrive-trigger/pull/103
but still not perfect. I've created an issue for this problem https://github.com/AbsaOSS/hyperdrive-trigger/issues/105

kevinwallimann (approved):

Issue #92, opened 2020-03-12
String filter for runs table 


jozefbakus:
String filter for run table.
Filter should contains input field used to filter data and trash can icon to filter removal. 
+ with backend call
String filter should be implemented on Workflow name and Project name column

Issue #89, opened 2020-03-12
Refresh runs table


jozefbakus:
Refresh runs table.
On refresh click should pull data from backend with set filters and sort.

PR #106, opened 2020-03-23
Feature/92 runs table string filter

jozefbakus:
implements #92
implements #89

### DIFF
--- a/ui/src/app/components/runs/filters/string-filter/string-filter.component.spec.ts
+++ b/ui/src/app/components/runs/filters/string-filter/string-filter.component.spec.ts
@@ -13,24 +13,32 @@
  * limitations under the License.
  */
 
-import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import {async, ComponentFixture, TestBed} from '@angular/core/testing';
 
-import { StringFilterComponent } from './string-filter.component';
+import {StringFilterComponent} from './string-filter.component';
+import {MockStore, provideMockStore} from "@ngrx/store/testing";
+import * as fromApp from "../../../../stores/app.reducers";
+import {Subject} from "rxjs";
 
 describe('StringFilterComponent', () => {
   let component: StringFilterComponent;
   let fixture: ComponentFixture<StringFilterComponent>;
 
+  const initialAppState = {};
+
   beforeEach(async(() => {
     TestBed.configureTestingModule({
-      declarations: [ StringFilterComponent ]
-    })
-    .compileComponents();
+      providers: [
+        provideMockStore({initialState: initialAppState})
+      ],
+      declarations: [StringFilterComponent]
+    }).compileComponents();
   }));
 
   beforeEach(() => {
     fixture = TestBed.createComponent(StringFilterComponent);
     component = fixture.componentInstance;
+    component.removeFiltersSubject = new Subject<any>();
     fixture.detectChanges();
   });
 
--- a/ui/src/app/components/runs/filters/string-filter/string-filter.component.spec.ts
+++ b/ui/src/app/components/runs/filters/string-filter/string-filter.component.spec.ts
@@ -1,0 +1,25 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+
+import { StringFilterComponent } from './string-filter.component';
+
+describe('StringFilterComponent', () => {
+  let component: StringFilterComponent;
+  let fixture: ComponentFixture<StringFilterComponent>;
+
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      declarations: [ StringFilterComponent ]
+    })
+    .compileComponents();
+  }));
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(StringFilterComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+});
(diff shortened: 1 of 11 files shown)


ui/src/app/components/runs/filters/string-filter/string-filter.component.spec.ts
```diff
@@ -0,0 +1,44 @@
+/*
```
kevinwallimann, line 1: Please add some tests. E.g. you could subscribe to the `changes` subject and call `modelChanged` or emit values to `removeFilterSubject`

ui/src/app/components/runs/runs.component.html
```diff
@@ -34,9 +34,17 @@
 
     <clr-dg-column [clrDgField]="dagRunColumns.WORKFLOW_NAME" [clrFilterValue]>
       <ng-container *clrDgHideableColumn="{hidden: false}">Workflow name</ng-container>
+      <clr-dg-filter [clrDgFilter]="workflowNameFilter">
+        <app-string-filter #workflowNameFilter [property]="dagRunColumns.WORKFLOW_NAME" [removeFiltersSubject]="removeFiltersSubject">
```
kevinwallimann, line 39: There is some duplication with `[property]` because it's already defined in the parent `clr-dg-column`. One way to resolve it is adding
```
  constructor(@Host() private column: ClrDatagridColumn) { }
```
to the `StringFilterComponent`. Then you can obtain the name from the column. However, obviously it adds a tight coupling to the `ClrDatagridColumn`, but one could argue that a filter without a column makes no sense

ui/src/app/components/runs/runs.component.ts
```diff
@@ -62,18 +69,52 @@ export class RunsComponent implements OnDestroy, AfterViewInit {
...
+      size: this.pageSize,
+      sort: this.sort,
+      filters: filters
     };
 
     this.store.dispatch(new GetDagRuns(searchRequestModel));
   }
 
+  createFiltersModel(filters: {[prop:string]: any}): FiltersModel {
+    let byWorkflowNameOption = this.filters[dagRunColumns.WORKFLOW_NAME];
```
kevinwallimann: You don't need `this.`

ui/src/app/components/runs/runs.component.ts
```diff
@@ -62,18 +69,52 @@ export class RunsComponent implements OnDestroy, AfterViewInit {
...
+    let filters:{[prop:string]: any} = {};
+    if (state.filters) {
+      for (let filter of state.filters) {
+        let {property, value} = <{property: string, value: any}>filter;
+        filters[property] = value;
+      }
+    }
+    this.filters = filters;
+
+    this.refresh();
```
kevinwallimann, line 80: You can pass filters directly into `this.refresh()`. Then you don't need `this.filters` at all.

ui/src/app/components/runs/filters/string-filter/string-filter.component.spec.ts
kevinwallimann: `accepts` also needs to be tested.

kevinwallimann (changes requested):

ui/src/app/components/runs/runs.component.ts
jozefbakus: Yes, I do need, because I call refresh method in view, where I don't have these values.

kevinwallimann: Ok I see

ui/src/app/components/runs/runs.component.ts
kevinwallimann: Can you mark the method private?

kevinwallimann: Also, consider using `const` instead of `let`

jozefbakus: Fixed

ui/src/app/components/runs/runs.component.html
jozefbakus: Fixed

jozefbakus: No, I've reverted it back. It is clearer this explicit way. It simplifies testing and I would like to avoid having references on parent elements from a child.

ui/src/app/components/runs/filters/string-filter/string-filter.component.spec.ts
jozefbakus: accepts has been tested

ui/src/app/components/runs/filters/string-filter/string-filter.component.ts
```diff
@@ -32,20 +32,18 @@ export class StringFilterComponent implements ClrDatagridFilterInterface<DagRunM
   //clarity interface
   changes = new Subject<any>();
   //angular
-  modelChanges: Subject<string> = new Subject<string>();
+  modelChanges: Subject<any> = new Subject<any>();
```
kevinwallimann: Why is this change required?

ui/src/app/components/runs/filters/string-filter/string-filter.component.ts
```diff
@@ -69,7 +67,7 @@ export class StringFilterComponent implements ClrDatagridFilterInterface<DagRunM
 
   onRemoveFilter() {
     this.value = undefined;
-    this.modelChanges.next(undefined)
+    this.modelChanges.next(this.value)
```
kevinwallimann, line 70: Why is this change needed?

jozefbakus: No reason. I just didnt want to repeat value.

kevinwallimann: Oh sorry, I completely missed `this.value = undefined;`

kevinwallimann (approved):

PR #108, opened 2020-03-25
Feature/92 generic filtering

### LINKED ISSUE #92: String filter for runs table

kevinwallimann:
Closes #92 

The goal of this PR is to provide a filtering implementation that can be reused across different columns and datagrids in the application.


ui/src/app/components/runs/runs.component.ts
```diff
@@ -67,45 +68,22 @@ export class RunsComponent implements OnDestroy, AfterViewInit {
...
   refresh() {
-    const filters: FiltersModel = this.createFiltersModel(this.filters);
-
-    let searchRequestModel: DagRunsSearchRequestModel = {
+    let searchRequestModel: TableSearchRequestModel = {
       from: this.pageFrom,
       size: this.pageSize,
       sort: this.sort,
-      filters: filters
+      containsFilterAttributes: this.filters.filter(f => f instanceof ContainsFilterAttributes)
```
jozefbakus, line 81: I like this.

jozefbakus (approved):
Love it!

Issue #94, opened 2020-03-12
Date filter for runs table


jozefbakus:
Date filter for runs table

- Range from - to open date interval filter (on ui use only date without time)
- Remove filter - trash can icon.
- Using backend.
- Used on started finished column

PR #110, opened 2020-03-26
Date range runs filter

jozefbakus:
implements: #94


kevinwallimann (commented):
image
1. On the calendar, when clicking one month back, the filter disappears. That should not happen.
2. The date text can not be edited manually. Is this on purpose? I would find it limiting

kevinwallimann (approved):

Issue #95, opened 2020-03-12
Status filter for runs table


jozefbakus:
Status filter for runs table.
- Status filter containing symbol + status text (multi-select or some similar component)
- Remove filter - trash can icon.
- Using backend.
- Used on status column

PR #111, opened 2020-03-26
Status runs filter

jozefbakus:
implements: #95

### DIFF
--- a/ui/src/app/components/runs/filters/status-filter/status-filter.component.ts
+++ b/ui/src/app/components/runs/filters/status-filter/status-filter.component.ts
@@ -1,0 +1,46 @@
+import {AfterViewInit, Component, Input, OnDestroy} from '@angular/core';
+import {Subject} from "rxjs";
+import {ClrDatagridFilterInterface} from "@clr/angular";
+import {DagRunModel} from "../../../../models/dagRuns/dagRun.model";
+
+@Component({
+  selector: 'app-status-filter',
+  templateUrl: './status-filter.component.html',
+  styleUrls: ['./status-filter.component.scss']
+}) export class StatusFilterComponent implements ClrDatagridFilterInterface<DagRunModel>, AfterViewInit, OnDestroy {
+  @Input() removeFiltersSubject: Subject<any>;
+  @Input() property: string;
+  @Input() statuses: {name: string, color: string, shape: string}[];
+  value: string = undefined;
+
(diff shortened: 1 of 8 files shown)


ui/src/app/components/runs/filters/status-filter/status-filter.component.ts
```diff
@@ -0,0 +1,46 @@
...
+
+@Component({
+  selector: 'app-status-filter',
+  templateUrl: './status-filter.component.html',
+  styleUrls: ['./status-filter.component.scss']
+}) export class StatusFilterComponent implements ClrDatagridFilterInterface<DagRunModel>, AfterViewInit, OnDestroy {
+  @Input() removeFiltersSubject: Subject<any>;
+  @Input() property: string;
+  @Input() statuses: {name: string, color: string, shape: string}[];
+  value: string = undefined;
```
kevinwallimann, line 15: Should the status filter not allow multiple selected statuses?

ui/src/app/components/runs/filters/status-filter/status-filter.component.ts
```diff
@@ -0,0 +1,46 @@
...
+import {DagRunModel} from "../../../../models/dagRuns/dagRun.model";
+
+@Component({
+  selector: 'app-status-filter',
+  templateUrl: './status-filter.component.html',
+  styleUrls: ['./status-filter.component.scss']
+}) export class StatusFilterComponent implements ClrDatagridFilterInterface<DagRunModel>, AfterViewInit, OnDestroy {
+  @Input() removeFiltersSubject: Subject<any>;
+  @Input() property: string;
+  @Input() statuses: {name: string, color: string, shape: string}[];
```
kevinwallimann: This tuple is worth an own interface imo, otherwise you'll have to edit multiple places if you change it

ui/src/app/components/runs/filters/status-filter/status-filter.component.ts
jozefbakus: No, I think that was my intention, but as a first step one value is fine

ui/src/app/components/runs/filters/status-filter/status-filter.component.ts
jozefbakus: fixed

kevinwallimann (approved):

Issue #115, opened 2020-03-27
Range filters should support open intervals


kevinwallimann:
Currently, the int range and datetime range filters fail if only either the start or only the end is given. Both start and end have to be passed.

With this issue, both the IntRangeFilter and DateTimeFilter should support open intervals. I.e. jobcounts greater than 2, all dag runs from 1st of february, all dag runs until march 23rd etc. At its core, this can be achieved e.g. as follows:

```
private def applyDateTimeRangeFilter(attributes: DateTimeRangeFilterAttributes, fieldMapping: Map[String, Rep[_]]): Rep[Boolean] = {
      val tableField = fieldMapping(attributes.field).asInstanceOf[Rep[LocalDateTime]]
      attributes.start.map(date => tableField >= date).getOrElse(LiteralColumn(true)) &&
        attributes.end.map(date => tableField <= date).getOrElse(LiteralColumn(true))
    }

```

Don't forget to add test cases!

PR #118, opened 2020-03-27
Range filters should support open intervals

jozefbakus:
implements: #115

### DIFF
--- a/src/test/scala/za/co/absa/hyperdrive/trigger/models/tables/SearchableTableQueryTest.scala
+++ b/src/test/scala/za/co/absa/hyperdrive/trigger/models/tables/SearchableTableQueryTest.scala
@@ -52,10 +52,10 @@ class SearchableTableQueryTest extends FlatSpec with Matchers with BeforeAndAfte
       ContainsFilterAttributes(field = stringField, value = "value"),
       ContainsFilterAttributes(field = stringField2, value = "str")
     ))
-    val intRangeFilterSeq = Some(Seq(IntRangeFilterAttributes(field = longField, start = 0, end = 1)))
+    val intRangeFilterSeq = Some(Seq(IntRangeFilterAttributes(field = longField, start = Option(0), end = Option(1))))
     val dateTimeRangeFilterSeq = Some(Seq(DateTimeRangeFilterAttributes(field = localDateTimeField,
-      start = LocalDateTime.of(2019, 1, 1, 1, 1, 1),
-      end = LocalDateTime.of(2021, 1, 1, 1, 1, 1))))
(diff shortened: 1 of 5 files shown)


src/test/scala/za/co/absa/hyperdrive/trigger/models/tables/SearchableTableQueryTest.scala
```diff
@@ -52,10 +52,10 @@ class SearchableTableQueryTest extends FlatSpec with Matchers with BeforeAndAfte
       ContainsFilterAttributes(field = stringField, value = "value"),
       ContainsFilterAttributes(field = stringField2, value = "str")
     ))
-    val intRangeFilterSeq = Some(Seq(IntRangeFilterAttributes(field = longField, start = 0, end = 1)))
+    val intRangeFilterSeq = Some(Seq(IntRangeFilterAttributes(field = longField, start = Option(0), end = Option(1))))
```
kevinwallimann, line 55: Why do you use `Option` instead of `Some`? It's shorter :) But no need to change ofc

src/test/scala/za/co/absa/hyperdrive/trigger/models/tables/SearchableTableQueryTest.scala
```diff
@@ -139,8 +139,28 @@ class SearchableTableQueryTest extends FlatSpec with Matchers with BeforeAndAfte
...
+
+    val result = await(db.run(underTest.search(searchRequest)))
+
+    val expected = Seq(TestSearchableData.t1, TestSearchableData.t3)
+    result.total should be > 0
+    result.total shouldBe expected.size
+    result.items should contain theSameElementsAs expected
+  }
+
+  "the date-time-range filter" should "find values greater or equal to the left bound" in {
```
kevinwallimann, line 161: What about values lower or equal to the right bound?

I would add that test, although it's kinda covered by that test below for the int range. Otherwise, please try to extract the common functionality in `applyIntRange` and `applyDateTimeRange`

src/test/scala/za/co/absa/hyperdrive/trigger/models/tables/SearchableTableQueryTest.scala
jozefbakus: In this case it doesnt matter. It handles nulls from java better.

src/test/scala/za/co/absa/hyperdrive/trigger/models/tables/SearchableTableQueryTest.scala
jozefbakus: It is covered by the test bellow. Done as you suggested.

kevinwallimann (approved):

PR #137, opened 2020-04-08
Feature/119 select multiple statuses for status filter

senelesithole:
1. New multi-status filter component.
2. Back-end implementation as well
3. Replaced double-quotes with single-quotes for front-end ( " " => ' : 'typescripts).
4. Replaced == with === ( == => === :typescripts).
5. Added semi-colon, end of line (  => ; :typescript).
6.Please tripple check the pom file before merge.


pom.xml
```diff
@@ -238,7 +239,7 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <version>${maven-surefire-plugin.version}</version>
                 <configuration>
-                    <skipTests>true</skipTests>
+                    <skipTests>${skipTests}</skipTests>
```
kevinwallimann: Why this change? The surefire plugin has to be disabled because we are using the scalatest plugin. The surefire plugin is automatically added by the maven default config

pom.xml
```diff
@@ -67,6 +67,7 @@
 
 
         <skip.docker>true</skip.docker>
+        <skipTests>true</skipTests>
```
kevinwallimann: I don't think tests should be skipped by default

src/main/resources/application.properties
```diff
@@ -61,9 +61,9 @@ sparkYarnSink.additionalConfs.spark.executor.memory=2g
 
 #Postgresql properties for connection to trigger metastore
 db.driver=org.postgresql.Driver
-db.url=jdbc:postgresql://
-db.user=
-db.password=
+db.url=jdbc:postgresql://localhost:5432/senele
+db.user=postgres
+db.password=pass
```
kevinwallimann: Be careful what you commit :)

src/main/scala/za/co/absa/hyperdrive/trigger/models/tables/SearchableTableQuery.scala
```diff
@@ -41,13 +41,23 @@ trait SearchableTableQuery {
...
         query.filter(table => applyDateTimeRangeFilter(attributes, table.fieldMapping)))
+      val withMultiContains = request.getContainsMultipleFilterAttributes.foldLeft(filteredQuery)((query, attributes) =>
+        query.filter((table => applyContainsMultipleFilter(attributes, table.fieldMapping))))
 
-      val length = filteredQuery.length.result
+//      val length = filteredQuery.length.result
+      val length = withMultiContains.length.result
 
-      val result = filteredQuery
+/*     val result = filteredQuery
```
kevinwallimann: Please don't commit commented code

src/main/scala/za/co/absa/hyperdrive/trigger/models/search/ContainsMultipleFilterAttributes.scala
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.hyperdrive.trigger.models.search
+
+case class ContainsMultipleFilterAttributes (override val field: String,
```
kevinwallimann: I think it should not be called `containsMultiple` but `equalsMultiple`. With `containsMultiple` I would expect a request like this:
```
{ 
...
"containsMultipleFilterAttributes":[{"field":"status","values":["Succeed"]}]
...
}
```
to return all the dag runs with status "Succeeded" because "Succeeded" contains "Succeed". However this is (correctly) not the case, an exact match is needed.

kevinwallimann: I would also remove everything related to the `StringEqualsFilter` because it's not used anywhere and is now just dead code. If we ever need such a filter, it can be built based on the `ContainsMultipleFilter`

src/main/scala/za/co/absa/hyperdrive/trigger/models/tables/SearchableTableQuery.scala
```diff
@@ -77,6 +87,11 @@ trait SearchableTableQuery {
       applyRangeFilter(tableField, attributes.start, attributes.end)
     }
 
+    private def applyContainsMultipleFilter(attributes: ContainsMultipleFilterAttributes, fieldMapping: Map[String, Rep[_]]): Rep[Boolean] = {
+      val tableField = fieldMapping(attributes.field).asInstanceOf[Rep[String]]
+      tableField inSetBind attributes.values
```
kevinwallimann, line 79: 👍 for using `inSetBind` instead of `inSet`

src/test/scala/za/co/absa/hyperdrive/trigger/models/tables/SearchableTableQueryTest.scala
```diff
@@ -243,6 +247,38 @@ class SearchableTableQueryTest extends FlatSpec with Matchers with BeforeAndAfte
...
+    val searchRequest = TableSearchRequest(
+      containsMultipleFilterAttributes = Some(Seq(filter)),
+      sort = None,
+      from = 0,
+      size = 50
+    )
+
+    val result = await(db.run(underTest.search(searchRequest)))
+
+    val expected = TestSearchableData.testSearchableEntities.filter(data => data.stringValue3.contains("foo") || data.stringValue3.contains("bar"))
```
kevinwallimann: Test says "values exactly equal" (which is correct), but here you're using `contains`

src/test/scala/za/co/absa/hyperdrive/trigger/models/tables/SearchableTableQueryTest.scala
```diff
@@ -243,6 +247,38 @@ class SearchableTableQueryTest extends FlatSpec with Matchers with BeforeAndAfte
...
+
+    result.total should be > 0
+    result.total shouldBe expected.size
+    result.items should contain theSameElementsAs expected
+  }
+
+  it should "not find values that do not contain the search string" in {
+    val filter = ContainsMultipleFilterAttributes(field = TestSearchableTableFieldNames.stringField3, values = List("not-matching-string", "not-matching-string2"))
```
kevinwallimann: In this negative test case, I would actually use strings like "fo" or "ar", to test for the fact that it really is an exact match and not just a contains

ui/package.json
```diff
@@ -38,22 +38,22 @@
...
     "karma-jasmine": "~2.0.1",
     "karma-jasmine-html-reporter": "^1.4.2",
     "protractor": "~5.4.3",
-    "ts-node": "~8.3.0",
+    "puppeteer": "~2.1.0",
+    "ts-node": "^8.3.0",
     "tslint": "~5.18.0",
-    "typescript": "~3.7.5",
-    "puppeteer": "~2.1.0"
+    "typescript": "~3.7.5"
```
kevinwallimann, line 57: Try not to commit changes like this. `ts-node` has changed from `~8.3.0` to `^8.3.0` (tilde to caret). I guess this was unintentional

ui/src/app/app-routing.module.ts
```diff
@@ -15,14 +15,14 @@
...
-import {LogInGuardService} from "./services/guards/logInGuard.service";
+import {HomeComponent} from './components/home/home.component';
+import {LoginComponent} from './components/login/login.component';
+import {WorkflowsComponent} from './components/workflows/workflows.component';
+import {RunsComponent} from './components/runs/runs.component';
+import {AuthGuardService} from './services/guards/authGuard.service';
+import {LogInGuardService} from './services/guards/logInGuard.service';
 import {routeNames} from './constants/routes.constants';
-import {WorkflowsHomeComponent} from "./components/workflows/workflows-home/workflows-home.component";
+import {WorkflowsHomeComponent} from './components/workflows/workflows-home/workflows-home.component';
```
kevinwallimann, line 25: Afaik, there is no technical argument for preferring either double-quotes or single quotes. One could make the argument to favor single quotes because it requires only one keystroke one a english keyboard while double quotes take to, but imo, that doesn't matter too much.
Eventually we will have to solve IDE auto-formatting problems by configuring an auto-formatter. For now I won't fuss much about it if your IDE does it automatically.

ui/src/app/app.module.ts
```diff
@@ -60,13 +61,15 @@ import {NumberRangeFilterComponent} from "./components/runs/filters/number-range
...
-    NumberRangeFilterComponent
+    NumberRangeFilterComponent,
+    MultipleStatusFilterComponent
   ],
   imports: [
     BrowserModule,
     AppRoutingModule,
     ClarityModule,
     FormsModule,
+    ReactiveFormsModule,
```
kevinwallimann: Do you really need the `ReactiveFormsModule`?

ui/src/app/components/runs/filters/datetime-range-filter/datetime-range-filter.component.ts
```diff
@@ -1,23 +1,23 @@
 /*
  * Copyright 2018 ABSA Group Limited
  *
- * Licensed under the Apache License, Version 2.0 (the "License");
+ * Licensed under the Apache License, Version 2.0 (the 'License');
```
kevinwallimann: Hmm did your IDE do that? That is very strange.

ui/src/app/components/runs/filters/datetime-range-filter/datetime-range-filter.component.ts
```diff
@@ -59,11 +59,11 @@ export class DatetimeRangeFilterComponent implements ClrDatagridFilterInterface<
   }
 
   isActive(): boolean {
-    return !!this.value.from || !!this.value.to
+    return !!this.value.from || !!this.value.to;
```
kevinwallimann, line 62: If your IDE did that ok. Otherwise, I wouldn't make changes to files unrelated to the issue, because it makes the review more time-consuming

ui/src/app/components/runs/filters/multiple-status-filter/multiple-status-filter.component.ts
```diff
@@ -0,0 +1,82 @@
...
+  changes = new Subject<any>();
+
+  constructor() {}
+
+  ngAfterViewInit(): void {
+    this.removeFiltersSubject.subscribe(_ => this.onRemoveFilter());
+  }
+
+  toggleStatuses(statusModel) {
+    if (!statusModel.checked) {
```
kevinwallimann: 1. If you really need the `checked` boolean, I would not store it directly on the `StatusModel`, but maybe create a map from `StatusModel` to boolean. `checked` only has a meaning within this component, but now spills over e.g. to `dagInstanceStatuses.constants.ts` where it has no meaning.

2. I think you don't need `checked` since it is equivalent with `selectedValues.indexOf(statusModel.name) != -1`. In fact, I think you should remove `checked` to eliminate any risk of inconsistency

ui/src/app/components/runs/filters/status-filter/status-filter.component.scss
```diff
@@ -1,6 +1,6 @@
 .status-selectable {
   cursor: pointer;
-  opacity: 0.3;
+  opacity: 0.5;
```
kevinwallimann: Why changing the status-filter? Also, I think the status-filter should be removed, because it's not used anywhere. If we will ever need it in the future, we can probably base it on the multiple-status-filter

ui/src/app/components/runs/filters/status-filter/status-filter.component.spec.ts
```diff
@@ -1,7 +1,22 @@
+/*
```
kevinwallimann: Remove the file

ui/src/app/components/runs/filters/status-filter/status-filter.component.ts
```diff
@@ -1,8 +1,23 @@
+/*
```
kevinwallimann: Remove the file

ui/src/app/components/runs/runs.component.html
```diff
@@ -61,22 +61,34 @@
...
+
+         <clr-dg-filter [clrDgFilter]="multiStatusFilter">
+        <app-multiple-status-filter #multiStatusFilter
+                                    [property]="dagRunColumns.STATUS"
+                                    [removeFiltersSubject]="removeFiltersSubject"
+                                    [statuses]="dagInstanceStatuses.getStatuses()">
+        </app-multiple-status-filter>
+      </clr-dg-filter>
+
+      <!-- <clr-dg-filter [clrDgFilter]="statusFilter">
```
kevinwallimann: Don't commit commented code

ui/src/app/components/runs/runs.component.ts
```diff
@@ -85,7 +86,8 @@ export class RunsComponent implements OnDestroy, AfterViewInit {
       stringEqualsFilterAttributes: this.filters.filter(f => f instanceof StringEqualsFilterAttributes),
```
kevinwallimann: Remove this line

ui/src/app/components/runs/runs.component.ts
```diff
@@ -96,7 +98,7 @@ export class RunsComponent implements OnDestroy, AfterViewInit {
   }
 
   clearSort() {
-    !!this.sort ? this.columns.find(_ => _.field == this.sort.by).sortOrder = 0 : undefined;
+    !!this.sort ? this.columns.find(_ => _.field === this.sort.by).sortOrder = 0 : undefined;
```
kevinwallimann, line 99: You are right about this one, but in the future please don't mix in code which is unrelated to the issue. It makes the review very time-consuming and one PR should deal with only one issue. Changing code unrelated to your issue also pose the risk of causing merge-conflicts with other open PRs.
It's better to open an issue for that which will be solved another time.

ui/src/app/models/search/stringEqualsFilterAttributes.model.ts
```diff
@@ -17,5 +17,6 @@ import {FilterAttributes} from './filterAttributes.model';
```
kevinwallimann, line 17: Remove this file

ui/src/app/stores/auth/auth.reducers.ts
```diff
@@ -13,13 +13,13 @@
...
 
-import * as AuthActions from "./auth.actions";
+import * as AuthActions from './auth.actions';
 import {localStorageKeys} from '../../constants/localStorage.constants';
 
 export interface State {
-  username: string,
-  isAuthenticated: boolean,
-  authenticationFailed: boolean
+  username: string;
```
kevinwallimann, line 20: What's bad about using commas instead of semicolons here?

kevinwallimann (changes requested):

src/test/scala/za/co/absa/hyperdrive/trigger/models/tables/SearchableTableQueryTest.scala
senelesithole: spot-on, fixed.

src/test/scala/za/co/absa/hyperdrive/trigger/models/tables/SearchableTableQueryTest.scala
senelesithole: I have changed it to == instead of .equals()
Let me know what you think

src/main/scala/za/co/absa/hyperdrive/trigger/models/tables/SearchableTableQuery.scala
senelesithole: Thank you 😊.

src/main/scala/za/co/absa/hyperdrive/trigger/models/search/ContainsMultipleFilterAttributes.scala
senelesithole: equalsMultiple makes more sense, fixed.
I will also do a single 'commit' where I removed **StringEqualsFilter** from the project as a whole.

src/main/scala/za/co/absa/hyperdrive/trigger/models/tables/SearchableTableQuery.scala
senelesithole: removed

src/main/resources/application.properties
senelesithole: my bad, thanks for pointing it out.

pom.xml
senelesithole: fixed (reverted) as well

pom.xml
senelesithole: I made this change from a hardcoded 'true' to using variable alias, anyways reverted back.

ui/package.json
senelesithole: how did that occur? :(  fixed.

ui/src/app/app-routing.module.ts
senelesithole: Thanks, I would like to keep it as single quotes.

ui/src/app/app.module.ts
senelesithole: For this PR no we do not, maybe for workflow screen we might need it. For now, I have removed it.

ui/src/app/components/runs/filters/datetime-range-filter/datetime-range-filter.component.ts
senelesithole: Yes, it did. find and replace in the wrong selection. Fixed.

ui/src/app/components/runs/filters/datetime-range-filter/datetime-range-filter.component.ts
senelesithole: I will start setting up a typescript plugin (if there is one) for IntelliJ instead of the VS Code.

ui/src/app/components/runs/filters/multiple-status-filter/multiple-status-filter.component.ts
senelesithole: Used the second suggestion, thanks.

ui/src/app/components/runs/filters/status-filter/status-filter.component.scss
senelesithole: Removed

ui/src/app/components/runs/filters/status-filter/status-filter.component.spec.ts
senelesithole: File removed.

ui/src/app/components/runs/filters/status-filter/status-filter.component.ts
senelesithole: File removed.

ui/src/app/components/runs/runs.component.html
senelesithole: Removed.

ui/src/app/components/runs/runs.component.ts
senelesithole: File removed.

ui/src/app/components/runs/runs.component.ts
senelesithole: noted with thanks :)

ui/src/app/models/search/stringEqualsFilterAttributes.model.ts
senelesithole: File removed.

ui/src/app/stores/auth/auth.reducers.ts
senelesithole: IDE: properties should be separated by semicolon, not commas.

pom.xml
```diff
@@ -29,7 +29,7 @@
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
         <scala.compat.version>2.11</scala.compat.version>
-        <spark.version>2.4.1</spark.version>
+        <spark.version>2.4.4</spark.version>
```
kevinwallimann: That isn't part of this issue

pom.xml
```diff
@@ -282,54 +277,7 @@
                     <noCache>true</noCache>
                 </configuration>
             </plugin>
-            <plugin>
```
kevinwallimann, line 285: You didn't remove this part on purpose, did you?

src/test/scala/za/co/absa/hyperdrive/trigger/models/tables/SearchableTableQueryTest.scala
```diff
@@ -243,6 +211,38 @@ class SearchableTableQueryTest extends FlatSpec with Matchers with BeforeAndAfte
...
+      sort = None,
+      from = 0,
+      size = 50
+    )
+
+    val result = await(db.run(underTest.search(searchRequest)))
+
+    result.total shouldBe 0
+  }
+
```
kevinwallimann, line 245: Perfect.

ui/src/app/components/runs/filters/multiple-status-filter/multiple-status-filter.component.ts
```diff
@@ -39,44 +39,41 @@ export class MultipleStatusFilterComponent implements ClrDatagridFilterInterface
...
-      }
+  ngOnDestroy(): void {
+    this.removeFiltersSubject.unsubscribe();
+  }
+
+toggleStatuses(statusModel) {
+  if (this.selectedValues.indexOf(statusModel.name) < 0) {
+    this.selectedValues = this.selectedValues.concat(statusModel.name);
+  } else {
+    if (this.selectedValues.indexOf(statusModel.name) >= 0) {
```
kevinwallimann: Do you need this if condition here? If you enter this else-block isn't `this.selectedValues.indexOf(statusModel.name)` always >= 0?

kevinwallimann: Also, please auto-format this file

kevinwallimann (changes requested):
Apart from the comments, this PR looks good to me now.
But I'm still feeling uneasy about all the formatting changes in unrelated files. First, they pose a risk for merge-conflicts. Second, let's say we notice some bug related to this PR in the future. It's going to cost a lot of time to filter out the changes that are actually relevant. I'd like to ask you to revert all the formatting changes to unrelated files and make a separate PR for them. I can help you with that if you want.

ui/src/app/stores/auth/auth.reducers.ts
kevinwallimann: ok

src/test/scala/za/co/absa/hyperdrive/trigger/models/tables/SearchableTableQueryTest.scala
kevinwallimann: Yes `==` is preferred over `.equals` in Scala, because it handles `null` properly. https://stackoverflow.com/questions/7681161/whats-the-difference-between-and-equals-in-scala

ui/src/app/components/runs/filters/datetime-range-filter/datetime-range-filter.component.ts
kevinwallimann: I've experimented with prettier, tslint and eslint here: 
https://github.com/AbsaOSS/hyperdrive-trigger/commit/d575ed6e308b4e381a78c7e174a53f6392908e99 and 
https://github.com/AbsaOSS/hyperdrive-trigger/commit/39396b59e77fc7cbc7c1e82584a842cbed8f96cc in one of the prototype branches for the new ui. We should definitely tackle this problem. Ideally, it would work both in IntelliJ and VSCode, if you prefer VSCode

senelesithole:
Okay, I'd like to set up some time with you tomorrow to revert back all the formatting changes ( the git way) rather than manually. 

Thanks a lot, Kevin!

Issue #116, opened 2020-03-27
Create abstract frontend filter


kevinwallimann:
Currently, `datetime-range-filter`, `number-range-filter`, `status-filter` and `string-filter` share common functionality. 

**Tasks**
- [x] Make filters generic, i.e. independent from `DagRunModel`
- [x] filters should be moved from `components/runs/filters` to e.g. `components/filters`

PR #139, opened 2020-04-16
Feature/119 select multiple statuses for status filter 2

senelesithole:
1. Adds a new multi-status-filer component.
2. Removes the string-filer component.


ui/src/app/components/runs/filters/multiple-status-filter/multiple-status-filter.component.ts
```diff
@@ -0,0 +1,83 @@
...
+  @Input() property: string;
+  @Input() statuses: StatusModel[];
+  selectedValues: string[] = [];
+  isSelected: any = [];
+
+  changes = new Subject<any>();
+
+  constructor() { }
+
+  ngAfterViewInit(): void {
```
kevinwallimann, line 43: Could you add this snippet here? I copied this from `datetime-range-filter.component.ts`. You'll have to adapt the usage of `modelChanges` as well. We will have to extract this common logic in #116 
```
    this.modelSubscription = this.modelChanges.pipe(
      debounceTime(500),
      distinctUntilChanged()
    ).subscribe(newValue => {
      this.changes.next();
    });
```
Otherwise, when you hit clear filters, you can see that there are two consecutive api calls, instead of only one

ui/src/app/components/runs/filters/multiple-status-filter/multiple-status-filter.component.ts
```diff
@@ -0,0 +1,83 @@
...
+    if (this.selectedValues.indexOf(statusModel.name) < 0) {
+      this.selectedValues = this.selectedValues.concat(statusModel.name);
+    } else {
+      this.selectedValues = this.selectedValues.filter(status => status !== statusModel.name);
+    }
+
+    this.changes.next(true);
+  }
+
+  onChange(i: number) {
```
kevinwallimann: Seems like this method doesn't do anything. Do you really need it?

kevinwallimann (changes requested):
This is a really clean PR!

ui/src/app/components/runs/filters/multiple-status-filter/multiple-status-filter.component.ts
senelesithole: Actually, I do not. I am already passing the index through [(ngModel)]="isSelected[i]".
Removed.

ui/src/app/components/runs/filters/multiple-status-filter/multiple-status-filter.component.ts
senelesithole: Added. 
Thanks, Kevin!

kevinwallimann (approved):

ui/src/app/components/runs/filters/multiple-status-filter/multiple-status-filter.component.ts
```diff
@@ -0,0 +1,90 @@
...
+      distinctUntilChanged()
+    ).subscribe(newValue => {
+      this.changes.next();
+    });
+
+    this.removeFiltersSubject.subscribe(_ => this.onRemoveFilter());
+  }
+
+  ngOnDestroy(): void {
+    this.removeFiltersSubject.unsubscribe();
```
jozefbakus, line 55: Could you please unsubscribe from modelSubscription here ?

ui/src/app/components/runs/filters/multiple-status-filter/multiple-status-filter.component.ts
```diff
@@ -0,0 +1,90 @@
...
+    if (this.selectedValues.indexOf(statusModel.name) < 0) {
+      this.selectedValues = this.selectedValues.concat(statusModel.name);
+    } else {
+      this.selectedValues = this.selectedValues.filter(status => status !== statusModel.name);
+    }
+
+    this.changes.next(true);
+  }
+
+  accepts(item: DagRunModel): boolean {
```
jozefbakus, line 69: This method should loop over selected statuses rather than over all statuses.
And I think, in case there is none selected status it should accept and return true.

jozefbakus: + update tests, because of logic change

jozefbakus (changes requested):

ui/src/app/components/runs/filters/multiple-status-filter/multiple-status-filter.component.ts
senelesithole: done

ui/src/app/components/runs/filters/multiple-status-filter/multiple-status-filter.component.ts
senelesithole: not so sure about my last test. let me know what you think.

jozefbakus: Hmmm logically it is still incorrect. You want in case of empty selectedValues return true.
In your solution if selectedValues is empty it never get into the cycle. 
`    if(this.selectedValues.length == 0) {
      return true;
    }
    for (const selectedStatus of this.selectedValues) {
      if(selectedStatus === item[this.property]) {
        return true;
      }
    }
    return false;`

senelesithole: Thanks :). Added as per suggestion.

jozefbakus (approved):
👍

senelesithole:
Multiple Status Filter impl back and front-end

Issue #132, opened 2020-03-30
Workflow create/update - workflow jobs part - Shell job type


jozefbakus:
Workflow create/update - workflow jobs part - Shell job type 

<img width="1645" alt="Screenshot 2020-04-28 at 16 01 15" src="https://user-images.githubusercontent.com/18142017/80496529-9d309680-8969-11ea-89bb-64c8dce26df8.png">

Issue #131, opened 2020-03-30
Workflow create/update - workflow jobs part - Spark job type


jozefbakus:
Workflow create/update - workflow jobs part - Spark job type

<img width="1645" alt="Screenshot 2020-04-28 at 16 01 15" src="https://user-images.githubusercontent.com/18142017/80496505-94d85b80-8969-11ea-8322-15f0f2fd749a.png">

Issue #130, opened 2020-03-30
Workflow create/update - workflow sensors part - Time sensor


jozefbakus:
Workflow create/update - workflow sensors part - Time sensor

<img width="1649" alt="Screenshot 2020-04-28 at 15 59 00" src="https://user-images.githubusercontent.com/18142017/80496369-678bad80-8969-11ea-9168-e3fbc0829dc7.png">

Issue #129, opened 2020-03-30
Workflow create/update - workflow sensors part - Absa Kafka sensor


jozefbakus:
Workflow create/update - workflow sensors part - Absa Kafka sensor

<img width="1649" alt="Screenshot 2020-04-28 at 15 59 00" src="https://user-images.githubusercontent.com/18142017/80496335-580c6480-8969-11ea-8950-b543bbf071d4.png">

PR #141, opened 2020-04-23
Feature/124 create edit workflow

jozefbakus:
Implements:
#124 
#130 
#129
#132 
#131 
#127 
#128


ui/src/app/stores/workflows/workflows.actions.ts
```diff
@@ -16,23 +16,106 @@
...
+import {WorkflowJoinedModel} from "../../models/workflowJoined.model";
+import {WorkflowFormPartsModel} from "../../models/workflowFormParts.model";
+import {WorkflowEntryModel} from "../../models/workflowEntry.model";
+import {JobEntryModel} from "../../models/jobEntry.model";
 
 export const INITIALIZE_WORKFLOWS = 'INITIALIZE_WORKFLOWS';
 export const INITIALIZE_WORKFLOWS_SUCCESS = 'INITIALIZE_WORKFLOWS_SUCCESS';
 export const INITIALIZE_WORKFLOWS_FAILURE = 'INITIALIZE_WORKFLOWS_FAILURE';
 
+export const STAR_WORKFLOW_INITIALIZATION = 'STAR_WORKFLOW_INITIALIZATION';
```
kevinwallimann: I guess this should be `START` instead of `STAR`

ui/src/app/stores/workflows/workflows.effects.ts
```diff
@@ -17,31 +17,128 @@ import {Injectable} from "@angular/core";
...
-            type: WorkflowActions.INITIALIZE_WORKFLOWS_FAILURE
+            type: WorkflowActions.LOAD_WORKFLOW_FAILURE_INCORRECT_ID
           }];
-        })
-      )})
+        } else {
+          return this.workflowService.getWorkflow(action.payload.id).pipe(
+
+            mergeMap((worfklow: WorkflowJoinedModel) => {
+              let workflowDetailsData = workflowFromPartsSequences.allDetails.map(detail => {
```
kevinwallimann: I would extract some parts into smaller methods. One for each `workflowDetailsData`, `sensorData` etc. Also try using `const` instead of `let` by default.

ui/src/app/stores/workflows/workflows.effects.ts
```diff
@@ -17,31 +17,128 @@ import {Injectable} from "@angular/core";
...
+          )
+        })
+      )
+    })
+  );
+
+  @Effect({dispatch: true})
+  workflowInitializationStart = this.actions.pipe(
+    ofType(WorkflowActions.STAR_WORKFLOW_INITIALIZATION),
+    withLatestFrom(
```
kevinwallimann, line 72: Why is this needed?

ui/src/app/stores/workflows/workflows.effects.ts
```diff
@@ -17,31 +17,128 @@ import {Injectable} from "@angular/core";
...
 export class WorkflowsEffects {
-  constructor(private actions: Actions, private workflowService: WorkflowService) {}
+  constructor(private actions: Actions, private workflowService: WorkflowService, private store: Store<AppState>) {}
 
   @Effect({dispatch: true})
   workflowsInitialize = this.actions.pipe(
     ofType(WorkflowActions.INITIALIZE_WORKFLOWS),
     switchMap((action: WorkflowActions.InitializeWorkflows) => {
       return this.workflowService.getProjects().pipe(
         mergeMap((projects: ProjectModel[]) => {
```
kevinwallimann: I would unnest these operators as much as possible. I.e. 
```
switchMap((action: WorkflowActions.InitializeWorkflows) => {
      return this.workflowService.getProjects()
}),
mergeMap((projects: ProjectModel[]) => {
```

ui/src/app/components/workflows/workflow/dynamic-parts/string-sequence-part/string-sequence-part.component.html
```diff
@@ -0,0 +1,29 @@
...
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<clr-control-container *ngFor="let val of value; let i = index; trackBy: trackByFn;">
+  <label *ngIf="i == 0">{{name}}</label>
+  <input clrControl [readOnly]="isShow" type="text" [size]="70" [ngModel]="val" (ngModelChange)="ngModelChanged($event, i)"/>
```
kevinwallimann, line 18: The size should be customizable. Some of the app arguments are cut off

ui/src/app/models/workflowEntry.model.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {WorkflowModel} from "./workflow.model";
+
+export class WorkflowEntryModel {
```
kevinwallimann, line 17: Maybe that should be renamed to DynamicFormEntryModel? It's not specific to the workflow

kevinwallimann (commented):
I really like the dynamic form components. They have a very clean abstraction. Each of those could have been a PR on its own. Also in general the code is very clean. Even though it seems pretty complicated, I was able to follow the logic.

However, I have to admit I completely underestimated the complexity of this degree of dynamic form generation. I'm a bit worried that little customizations to the form might become quite hard to add in the future, e.g. async check if workflow name is unique. In fact, where we need the dynamic forms the most is imo for the app arguments. And there it seems almost impossible now to have dynamic form elements instead of just a set of string elements. This would require nested dynamic forms and I think that's definitely too complex.

Zejnilovic:
> Implements:
> #124
> #130
> #129
> #132
> #131
> #127
> #128

Just a small heads up. If you want GitHub to automatically close all of these just write `Fixes` in front of each of the numbers.

ui/src/app/models/workflowFormParts.model.ts
```diff
@@ -0,0 +1,50 @@
...
+ * limitations under the License.
+ */
+
+export class WorkflowFormPartsModel {
+  constructor(
+    public detailsParts: FormPart[],
+    public dynamicSwitchSensorPart: FormPart,
+    public staticJobPart: FormPart,
+    public dynamicSwitchJobPart: FormPart,
+    public dynamicParts: DynamicFormParts
```
kevinwallimann, line 22: 1. `dynamicSwitchSensorPart` -> `sensorSwitch` or `sensorSwitchPart`? The switch itself is not dynamic so I would omit the word dynamic. Same for `dynamicSwitchJobPart`
2. How about including `sensorDynamicParts` and `jobDynamicParts` directly here without the extra hierarchy of `DynamicFormParts`?

ui/src/app/services/workflow/workflow.service.ts
```diff
@@ -33,4 +35,53 @@ export class WorkflowService {
...
 
+  getWorkflow(id: number): Observable<WorkflowJoinedModel> {
+    let params = new HttpParams().set('id', id.toString());
+
+    return this.httpClient.get<WorkflowJoinedModel>(api.GET_WORKFLOW, {params: params, observe: 'response'}).pipe(
+      map(response => response.body)
+    );
+  }
+
+  getWorkflowComponents(): Observable<DynamicFormParts> {
```
kevinwallimann: Maybe `getWorkflowComponents` -> `getWorkflowDynamicFormParts`? When I read workflow components, I originally expected to get details, sensor and jobs, which confused me in the workflow-effects for `INITIALIZE_WORKFLOWS`

ui/src/app/stores/workflows/workflows.effects.ts
```diff
@@ -17,31 +17,128 @@ import {Injectable} from "@angular/core";
...
+                workflowFormPartsConsts.SENSOR.SENSOR_TYPE,
+                workflowFormPartsConsts.JOB.JOB_NAME,
+                workflowFormPartsConsts.JOB.JOB_TYPE,
+                workflowComponents
+              );
+              return [{
+                type: WorkflowActions.INITIALIZE_WORKFLOWS_SUCCESS,
+                payload: {
+                  projects: projects,
+                  workflows: [].concat(...projects.map((project) => project.workflows)),
```
kevinwallimann: Not sure if it's a good idea to define `workflows` as part of the state. It could lead to inconsistencies with `projects`. I would do the transformation where the value is consumed.

ui/src/app/stores/workflows/workflows.reducers.ts
```diff
@@ -16,27 +16,141 @@
...
   projects: ProjectModel[],
-  workflows: WorkflowModel[]
-  loading: boolean
+  workflows: WorkflowModel[],
+  loading: boolean,
+  workflowAction: {
+    id: number,
+    mode: string,
+    loading: boolean,
+    workflow: WorkflowJoinedModel,
```
kevinwallimann, line 29: Where is this used?

ui/src/app/stores/workflows/workflows.reducers.ts
```diff
@@ -16,27 +16,141 @@
...
+    workflowData: {
+      details: {property: string, value: any}[],
```
kevinwallimann, line 31: This can be typed with `WorkflowEntryModel[]` Same for `sensor` and `jobs`(`JobEntryModel`)

ui/src/app/stores/workflows/workflows.effects.ts
kevinwallimann: In principle, this section describes a mapping from `WorkflowJoinedModel` to `workflowData`. How about defining a type for `workflowData` to make this fact more explicit? Then the whole mapping can be extracted to a separate class.

ui/src/app/stores/workflows/workflows.effects.ts
```diff
@@ -17,31 +17,128 @@ import {Injectable} from "@angular/core";
...
+
+            mergeMap((worfklow: WorkflowJoinedModel) => {
+              let workflowDetailsData = workflowFromPartsSequences.allDetails.map(detail => {
+                let value = get(worfklow, detail.property);
+                if(value != undefined) {
+                  return new WorkflowEntryModel(detail.property, value);
+                }
+              });
+
+              let sensorDynamicPart = workflowFormPartsConsts.SENSOR.SENSOR_TYPE;
```
kevinwallimann: I would just name it `sensorType`. In this context `sensorDynamicPart` is misleading imo because it suggest to be a single element of `sensorDynamicParts` just below, which is not the case.

ui/src/app/stores/workflows/workflows.reducers.ts
```diff
@@ -16,27 +16,141 @@
...
+      const emptyJobData = {order: action.payload, job: []};
+      const jobs = [...state.workflowAction.workflowData.jobs, emptyJobData];
+      return {...state, workflowAction: {
+          ...state.workflowAction, workflowData: {
+            ...state.workflowAction.workflowData, jobs: [...jobs]
+          }
+        }};
+    case (WorkflowsActions.WORKFLOW_JOB_CHANGED):
+      const oldJobDataOption = state.workflowAction.workflowData.jobs.find(job => job.order == action.payload.order);
+      const oldJobData: {property: string, value: any}[] = !!oldJobDataOption ? oldJobDataOption.job : [];
```
kevinwallimann: Can be precisely typed with `WorkflowEntryModel[]`

ui/src/app/stores/workflows/workflows.reducers.ts
```diff
@@ -16,27 +16,141 @@
...
+      return {...state, workflowAction: {
+          ...initialState.workflowAction, loading: false
+        }};
+    case (WorkflowsActions.LOAD_WORKFLOW_FAILURE_INCORRECT_ID):
+      return {...state, workflowAction: {
+          ...initialState.workflowAction, loading: false
+        }};
+    case (WorkflowsActions.WORKFLOW_DETAILS_CHANGED):
+      const detailsData = [
+        ...state.workflowAction.workflowData.details.filter(item => item.property !== action.payload.property),
```
kevinwallimann, line 89: Why do you need to explicitly filter out the changed property. Is it not overwritten by `action.payload`?

ui/src/app/stores/workflows/workflows.reducers.ts
```diff
@@ -16,27 +16,141 @@
...
+      const updatedJobsData = [
+        ...state.workflowAction.workflowData.jobs.filter(jobEntries => jobEntries.order !== action.payload.order),
+        {order: action.payload.order, job: updatedJobData}
+      ];
+      return {...state, workflowAction: {
+          ...state.workflowAction, workflowData: {
+            ...state.workflowAction.workflowData, jobs: [...initialState.workflowAction.workflowData.jobs, ...updatedJobsData]
+          }
+        }};
+    case (WorkflowsActions.WORKFLOW_JOB_CLEANED):
```
kevinwallimann: I think I would rename this to `WORKFLOW_JOB_TYPE_SWITCHED`. Imo, it's not really a clean because you still need to keep the value of the job type. Moreover, cleaning even the job name when switching the job type is not expected behavior imo, but this can be addressed in the future

kevinwallimann: The same argument doesn't exactly hold for `WORKFLOW_SENSOR_CLEANED`, but for the sake of consistency I would change it as well.

ui/src/polyfills.ts
```diff
@@ -70,9 +70,14 @@
 /***************************************************************************************************
  * Zone JS is required by default for Angular itself.
  */
-import 'zone.js/dist/zone';  // Included with Angular CLI.
-
+import 'zone.js/dist/zone';
+import {environment} from "./environments/environment";  // Included with Angular CLI.
 
+if(!environment.production){
+  (Map.prototype as any).toJSON = function () {
```
kevinwallimann, line 77: Why is this needed?

ui/src/app/components/workflows/workflow/dynamic-parts/cron-quartz-part/cron-quartz-part.component.ts
```diff
@@ -0,0 +1,213 @@
+import {AfterViewInit, Component, Input, OnInit} from '@angular/core';
```
kevinwallimann, line 1: You said this component needs more work, so I won't review it here. Can be completed in a separate PR.

ui/src/app/components/workflows/workflow/dynamic-parts/key-string-value-part/key-string-value-part.component.spec.ts
```diff
@@ -0,0 +1,40 @@
+/*
```
kevinwallimann, line 1: I think this component needs some non-trivial testing, but can be done in a different PR

ui/src/app/constants/workflowFromParts.constants.ts
```diff
@@ -0,0 +1,37 @@
...
+  SENSOR: {
+    SENSOR_TYPE: new FormPart('Sensor type', 'sensorType.name', true, 'select-field'),
+  },
+  JOB: {
+    JOB_NAME: new FormPart('Job name', 'name', true, 'string-field'),
+    JOB_TYPE: new FormPart('Job type', 'jobType.name', true, 'select-field')
+  }
+};
+
+export const workflowFromPartsSequences = {
```
kevinwallimann: Typo `From` -> `Form`. Also for the filename itself

ui/src/app/components/workflows/workflow/dynamic-parts/key-string-value-part/key-string-value-part.component.spec.ts
jozefbakus: Yes, https://github.com/AbsaOSS/hyperdrive-trigger/issues/143

ui/src/app/components/workflows/workflow/dynamic-parts/cron-quartz-part/cron-quartz-part.component.ts
jozefbakus: Yes, https://github.com/AbsaOSS/hyperdrive-trigger/issues/145

ui/src/polyfills.ts
jozefbakus: It is only for development because of !environment.production.
It is for Redux dev tools chrome plugin that I can see map content. 
https://chrome.google.com/webstore/detail/redux-devtools/lmhkpmbekcpmknklioeibfkpmmfibljd?hl=en

ui/src/app/constants/workflowFromParts.constants.ts
jozefbakus: Fixed

ui/src/app/stores/workflows/workflows.reducers.ts
jozefbakus: I've changed that. I think job cleaning is fine, I did it intentionally. But we can discuss it during next planing session.

ui/src/app/stores/workflows/workflows.reducers.ts
jozefbakus: No. state.workflowAction.workflowData.details is array and action.payload is one element so you have to do it otherwise it won't override but add an element to the array. You can end up with duplicates.

ui/src/app/stores/workflows/workflows.reducers.ts
jozefbakus: Fixed

ui/src/app/stores/workflows/workflows.effects.ts
jozefbakus: Done

ui/src/app/stores/workflows/workflows.reducers.ts
jozefbakus: No, I dont want to do it at least here. Reducer should not involve any models. It is easier to create empty state. In case of WorkflowJoinedModel there is no other way but for data I there is.

ui/src/app/stores/workflows/workflows.reducers.ts
jozefbakus: Nowhere. But I think I will use that on form submit. If no then I will remove it.

ui/src/app/services/workflow/workflow.service.ts
jozefbakus: Fixed

ui/src/app/models/workflowFormParts.model.ts
jozefbakus: 1. fixed.
2. maybe when we have backend endpoints for dynamic parts. Right now it is hardcoded so it is simpler to load it into one model.

ui/src/app/models/workflowEntry.model.ts
jozefbakus: Yes, but I would leave it as it is. It is used only in workflow form so that is why I used that name. DynamicFormEntryModel doesn't make sense for me because it is used also in non-dynamic parts. It could be renamed to FormEntryModel, but I would leave it as it is.

ui/src/app/components/workflows/workflow/dynamic-parts/string-sequence-part/string-sequence-part.component.html
jozefbakus: Yes, I agree. Let's do it in the next sprint.

ui/src/app/stores/workflows/workflows.effects.ts
jozefbakus: fixed

ui/src/app/stores/workflows/workflows.effects.ts
jozefbakus: Provide the last value from another observable.
So it uses the latest store state.

ui/src/app/stores/workflows/workflows.actions.ts
jozefbakus: Fixed

ui/src/app/stores/workflows/workflows.effects.ts
jozefbakus: Fixed

ui/src/app/stores/workflows/workflows.reducers.ts
kevinwallimann: Hmm ok

ui/src/app/stores/workflows/workflows.effects.ts
jozefbakus: good idea. fixed

ui/src/app/stores/workflows/workflows.reducers.ts
kevinwallimann: My concern is that when `WorkflowEntryModel` changes, you will have to change the type of `details` as well. But your concern of not involving any models is also valid. So I leave it up to you

ui/src/app/constants/workflowFromParts.constants.ts
```diff
@@ -0,0 +1,37 @@
+/*
```
kevinwallimann: The filename still has the typo

kevinwallimann (changes requested):

ui/src/app/constants/workflowFromParts.constants.ts
jozefbakus: done

kevinwallimann (approved):

PR #153, opened 2020-05-12
Feature/145 time sensor final touch

### LINKED ISSUE #145: Time sensor final touch
Cron-quartz part: 

**SHOW**

1. Show valid: 
If **isShow and cron-quartz expression (Input value) is valid** then user should see human readable text representation of cron quartz.
Screenshot 2020-05-20 at 08 13 12

2. Show invalid: 
If **isShow and cron-quartz expression (Input value) is not valid** then user instead text representation of cron quartz should see text with "_Invalid cron expression, please update your workflow with correct cron expression_" (or some similar text in red)
Screenshot 2020-05-20 at 08 20 31

**EDIT**
Note: design of this component is up to implementor.

1. Switch
 User should be able switch between free text cron input and user friendly cron input (as it is now).

2. Switch free text to user friendly input: 
If it possible to switch it and use value from free text then pre set selects. 
If it is not possible then show confirmation dialog (with info that it is not possible and default value will be used) and use default value. (You can check how confirmation dialog used on other places).

3. Switch user friendly input to free text: 
Cron expression set in user friendly input should be used in free text input. 

4. When user goes from show workflow to edit workflow, user friendly input should be preferred but if it is not possible, free text input should be used. 

5. When user goes from show workflow to edit workflow and cron expression is invalid, toast with this information should be shown to user and user should be able update invalid cron expression in free text input.

NOTE: You can look for javascript lib which will transform cron quartz expression to human readable text

Design proposal: 

- View mode: 
Screenshot 2020-05-25 at 08 05 32

- Free text cron quartz expression - edit mode:  
Screenshot 2020-05-25 at 08 17 23

- User friendly input - edit mode: 
Screenshot 2020-05-25 at 08 12 39

senelesithole:
Creating a workflow using Cron sensor type.

translate view expression to Cron expression, front-end.
fixes #145


ui/src/app/constants/cronExpressionOptions.constants.ts
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+ export const cronExpressionOptions = {
+
+    MINUTES_VALUES : [5, 10, 15, 20, 25, 30, 35, 40, 45, 50, 55],
+    MINUTE_VALUES : [0, 5, 10, 15, 20, 25, 30, 35, 40, 45, 50, 55],
+    HOUR_VALUES : [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23],
+    FREQUENCIES : [
```
kevinwallimann: I think you should use an enum here. Then you don't have to operate with the numbers, which is hard to maintain

ui/src/app/constants/cronExpressionOptions.constants.ts
```diff
@@ -0,0 +1,35 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+ export const cronExpressionOptions = {
+
+    MINUTES_VALUES : [5, 10, 15, 20, 25, 30, 35, 40, 45, 50, 55],
```
kevinwallimann: Not sure if we should include the values > 30. They cover rather unusual schedules like 00:00, 00:40, 01:00, 01:40 etc. At first sight, it's hard to grasp the difference between every hour every and every hour at, because the dropdown is almost the same.

ui/src/app/components/workflows/workflow/dynamic-parts/cron-quartz-part/cron-quartz-part.component.ts
```diff
@@ -16,198 +32,73 @@ export class CronQuartzPartComponent implements OnInit, AfterViewInit {
 
   base;
```
kevinwallimann: Please add types to the variables

ui/src/app/components/workflows/workflow/dynamic-parts/cron-quartz-part/cron-quartz-part.component.ts
```diff
@@ -16,198 +32,73 @@ export class CronQuartzPartComponent implements OnInit, AfterViewInit {
...
-    }
-
-    if (n && n.base && n.base === 4) {
-      cron[3] = '?';
-      cron[5] = n.dayValue;
-    }
-
-    if (n && n.base && n.base >= 5) {
-      cron[3] = typeof n.dayOfMonthValue !== undefined ? n.dayOfMonthValue : '?';
+  setCron(value, label) {
```
kevinwallimann: I think extracting the logic to this method doesn't really reduce complexity. You could write it in the calling method like this:

```
this.minuteValue = option;
const cronMinute = `0/${value}`;
this.cron = ['0', cronMinute, '0', '?', '*', '*', '*'];
this.modelChanged();
```

So here you reuse only `this.modelChanged()` at while introducing a switch-case and a default-branch which is never used.

ui/src/app/components/workflows/workflow/dynamic-parts/cron-quartz-part/cron-quartz-part.component.ts
```diff
@@ -16,198 +32,73 @@ export class CronQuartzPartComponent implements OnInit, AfterViewInit {
...
-    return cron.join(' ');
+    this.modelChanged();
   }
 
   fromCron(value: string) {
     const cron: string[] = value.replace(/\s+/g, ' ').split(' ');
-    // var frequency = {base: '1'}; // default: every minute
 
     if (cron[1] === '*' && cron[2] === '*' && cron[3] === '*' && cron[4] === '*' && cron[5] === '?') {
       this.base = 1; // every minute
```
kevinwallimann: Please use an enum instead of numbers

ui/src/app/constants/cronExpressionOptions.constants.ts
```diff
@@ -0,0 +1,35 @@
...
+    MINUTE_VALUES : [0, 5, 10, 15, 20, 25, 30, 35, 40, 45, 50, 55],
+    HOUR_VALUES : [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23],
```
kevinwallimann: On the UI, this will be "Every day at 3". Imo, this isn't very intuitive. As a user, I would expect "Every day at 03:00", so you might need a map here

ui/src/app/components/workflows/workflow/dynamic-parts/cron-quartz-part/cron-quartz-part.component.ts
```diff
@@ -16,198 +32,73 @@ export class CronQuartzPartComponent implements OnInit, AfterViewInit {
...
     }
-
-    if (n && n.base && n.base === 6) {
-      cron[4] = typeof n.monthValue !== undefined ? n.monthValue : '*';
-    }
-    return cron.join(' ');
+    this.modelChanged();
   }
 
   fromCron(value: string) {
```
kevinwallimann, line 127: This code is never called. I think you should use something like this to fill the values for `minuteValue`, `hourValue`, `dayValue` when the input value changes. Btw how to intercept input property: https://angular.io/guide/component-interaction#intercept-input-property-changes-with-a-setter

ui/src/app/components/workflows/workflow/dynamic-parts/cron-quartz-part/cron-quartz-part.component.html
```diff
@@ -67,6 +57,6 @@
 <div *ngIf="isShow">
   <clr-control-container>
     <label>{{name}}</label>
-    <input clrControl type="text" [readOnly]="true"  name="workflowName" [size]="40" value="Every day at 12:00" />
+    <input clrControl type="text" [readOnly]="true" name="workflowName" [size]="40" value="Every day at 12:00" />
```
kevinwallimann: This should not be a constant value

ui/src/app/components/workflows/workflow/dynamic-parts/cron-quartz-part/cron-quartz-part.component.spec.ts
```diff
@@ -1,25 +1,83 @@
+/*
```
kevinwallimann, line 1: Nice tests 👍

kevinwallimann (changes requested):

ui/src/app/constants/cronExpressionOptions.constants.ts
senelesithole: fixed

ui/src/app/constants/cronExpressionOptions.constants.ts
senelesithole: not so sure about the way I implemented enums.

ui/src/app/components/workflows/workflow/dynamic-parts/cron-quartz-part/cron-quartz-part.component.ts
senelesithole: fixed

ui/src/app/components/workflows/workflow/dynamic-parts/cron-quartz-part/cron-quartz-part.component.ts
senelesithole: thanks for this. eye-opener.

ui/src/app/components/workflows/workflow/dynamic-parts/cron-quartz-part/cron-quartz-part.component.ts
senelesithole: fixed

ui/src/app/constants/cronExpressionOptions.constants.ts
senelesithole: fixed

ui/src/app/components/workflows/workflow/dynamic-parts/cron-quartz-part/cron-quartz-part.component.ts
senelesithole: the concept around this method is getting a Cron expression from the back-end, do some manipulations and then show it to the UI ( the way I understood it ). Leaving it as-is for now.

ui/src/app/components/workflows/workflow/dynamic-parts/cron-quartz-part/cron-quartz-part.component.spec.ts
senelesithole: thanks 😊.

jozefbakus:
It is nice, I would do couple of changes: 
1.  During workflow creation, I would set some default value. (so in ngOnInit - check if value is empty, and if yes then emit default value)
2. For whole time value should be valid. If you change frequency (day, hour at, hour every) rest of the visible fields should be set to default value. (So in any point you will have valid cron expression) 
3. In show mode I would show same components as during edit but without possibility to update. (You can use [readOnly]="isShow" input property but not all fields should be visible when it doesnt make sense)
4. fromCron is there for ngOnInit. When component instance is created it should take input value if it is not empty and translate it to your cron field and pre set everything you need)
5. If input value is not empty and it contains incorrect cron expression I would set default values and show toast with information that it is not correct and default value is used. (for toast we use toastrService implemented in https://github.com/AbsaOSS/hyperdrive-trigger/pull/154/files)

ui/src/app/components/workflows/workflow/dynamic-parts/cron-quartz-part/cron-quartz-part.component.html
```diff
@@ -67,6 +59,15 @@
 <div *ngIf="isShow">
```
jozefbakus, line 81: I would repleace this part with: 

`  <clr-control-container>
    <label>{{name}}</label>
    <input *ngIf="showGuard == 1" [readOnly]="true" clrControl value="Every {{ showFrequency }} {{ cronValue }} minutes"
           [size]="40" />
    <input *ngIf="showGuard == 2" [readOnly]="true" clrControl
           value="Every {{ showFrequency }} {{ cronValue }} minutes past the hour" [size]="40" />
    <input *ngIf="showGuard == 3" [readOnly]="true" clrControl value="Every {{ showFrequency }} at {{ cronValue }}:00"
           [size]="40" />
  </clr-control-container>
`

It will appear the same way as the rest of the parts.

jozefbakus (changes requested):
1. If mode is show that its ok, but when you edit everything is empty. You dont have to use variables: 
  showFrequency, cronValue, showGuard. You can reuse base, dayValue, minuteValue, hourValue for both modes. 

2.  You need some method which will say whether cron expression is valid. 
 You can pull changes from feature/new-ui and use some parts from there for toasts. 
Two use cases: 
A. Is show and cron expression is invalid show toast that it is invalid and show it in view that it is invalid. 
B. Is edit  and cron expression is invalid show toast with information that you will change value to default

ui/src/app/components/workflows/workflow/dynamic-parts/cron-quartz-part/cron-quartz-part.component.html
kevinwallimann: Btw @jozefbakus if you use three backticks, the linebreaks will be preserved in the comment

ui/src/app/constants/cronExpressionOptions.constants.ts
kevinwallimann: 👍 for the map, but I would expect "03" (with a leading zero)

ui/src/app/components/workflows/workflow/dynamic-parts/cron-quartz-part/cron-quartz-part.component.ts
kevinwallimann: Hmm... In `cron-quartz-component.html` you still have code like `base == 1` everywhere. After three months you will have forgotten what that 1 stands for and where it comes from.

jozefbakus (changes requested):
1. I miss this 
"When user goes from show workflow to edit workflow, user friendly input should be preferred but if it is not possible, free text input should be used."
When I have workflow with "0,17,23,41 0 0 ? * * *" and I click on edit user friendly part is rendered (but free text should be used)


2. When I use 0 10 * ? * * *asd, it says it is valid expression. 
I think for validation and translation you can use backend endpoint "/util/quartzDetail"

kevinwallimann:
2 problems in the free-text expression
I save a workflow with free-text expression */10 * * ? * *
image

1. In the show mode, no text is shown
image

2. Back in the edit mode, the drop down is selected on User-friendly input
image

ui/src/app/components/workflows/workflow/dynamic-parts/cron-quartz-part/cron-quartz-part.component.ts
```diff
@@ -128,28 +131,17 @@ export class CronQuartzPartComponent implements OnInit {
     if (showCron[1] !== '*' && isNaN(+showCron[1])) {
```
kevinwallimann: The if-conditions in this method still need to be much more precise.
Example: The every-hour-every expression has to look like this:
`0 0/{x} * ? * * *`, where `{x}` can be one of  5, 10, 15, 20, 25 or 30. That's how you generate it in `onMinuteSelect`. Conceptually, the above if-condition should be like this:
`
if (showCron == '0 0/5 * ? * * *' || showCron == '0 0/10 * ? * * *' || showCron == '0 0/15 * ? * * *' || showCron == '0 0/20 * ? * * *' || showCron == '0 0/25 * ? * * *' || showCron == '0 0/30 * ? * * *')
`
Of course, you can make it better readable, reusable etc. Also for the other two if conditions, think about what are all the possible cron expressions that can be generated from the "Every hour at" and "Every Day at" options. Any cron-expression which cannot be created with the user-friendly input has to be displayed as free-text

Issue #148, opened 2020-04-28
Cancel during create/update/copy workflow


jozefbakus:
Cancel during create/update/copy workflow.
On click, a user should be redirected to the previous page

<img width="1650" alt="Screenshot 2020-04-28 at 16 27 18" src="https://user-images.githubusercontent.com/18142017/80499477-517fec00-896d-11ea-8001-7accf85d7576.png">

PR #156, opened 2020-05-13
Feature/148 cancel workflow during action

jozefbakus:
implement: #148

### DIFF
--- a/ui/src/app/components/workflows/workflow/workflow.component.ts
+++ b/ui/src/app/components/workflows/workflow/workflow.component.ts
@@ -14,13 +14,15 @@
  */
 
 import { Component, OnDestroy, OnInit } from '@angular/core';
-import { ActivatedRoute } from '@angular/router';
+import { ActivatedRoute, Router } from '@angular/router';
 import { AppState, selectWorkflowState } from '../../../stores/app.reducers';
 import { Subscription } from 'rxjs';
 import { Store } from '@ngrx/store';
 import { StartWorkflowInitialization } from '../../../stores/workflows/workflows.actions';
 import { workflowModes } from '../../../models/enums/workflowModes.constants';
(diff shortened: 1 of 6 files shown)


senelesithole (approved):
Nice and simple to follow.

ui/src/app/components/workflows/workflow/workflow.component.ts
```diff
@@ -68,6 +76,10 @@ export class WorkflowComponent implements OnInit, OnDestroy {
     this.isJobsAccordionHidden = !this.isJobsAccordionHidden;
   }
 
+  cancelWorkflow() {
+    !!this.previousRouteService.getPreviousUrl() ? this.location.back() : this.router.navigateByUrl(absoluteRoutes.WORKFLOWS_HOME);
```
kevinwallimann: When the user clears the history while he is on the workflow page, does it work then? Why can't we just call `navigateByUrl(getPreviousUrl())`?

ui/src/app/services/previousRoute/previous-route.service.spec.ts
```diff
@@ -0,0 +1,55 @@
+/*
```
kevinwallimann, line 1: Nice! 👍

ui/src/app/components/workflows/workflow/workflow.component.ts
jozefbakus: It will work because history is saved in our service (even if that is empy it will go to . WORKFLOWS_HOME page).
Because history can be empty and you dont wont to redirect user to blank browser page but to something in your app (in this case it is WORKFLOWS_HOME page) .

jozefbakus: After discussion, yes you are right. Fixed.

kevinwallimann (approved):

Issue #149, opened 2020-04-28
On create/update user redirect


jozefbakus:
On create/update user should be redirected to show workflow page. 

<img width="1634" alt="Screenshot 2020-04-28 at 16 31 34" src="https://user-images.githubusercontent.com/18142017/80500020-f3073d80-896d-11ea-9c03-1d52178a814e.png">
<img width="1680" alt="Screenshot 2020-04-28 at 16 30 34" src="https://user-images.githubusercontent.com/18142017/80500024-f4386a80-896d-11ea-9e86-39cb38b4b18d.png">

Issue #126, opened 2020-03-30
Integration of update/create workflow screen with corresponding backend endpoint 


jozefbakus:
Integration of update/create workflow screen with corresponding backend endpoint.
Call for update/create workflow + update state.

+ update buttons submitting buttons to 'Save'

PR #159, opened 2020-05-18
Feature/126 create update backend integration

jozefbakus:
Implements: #126 and #149


ui/src/app/components/workflows/workflow/workflow.component.ts
```diff
@@ -121,6 +123,22 @@ export class WorkflowComponent implements OnInit, OnDestroy {
...
 
+  createWorkflow() {
+    this.confirmationDialogServiceSubscription = this.confirmationDialogService
+      .confirm(ConfirmationDialogTypes.YesOrNo, texts.CREATE_WORKFLOW_CONFIRMATION_TITLE, texts.CREATE_WORKFLOW_CONFIRMATION_CONTENT)
+      .subscribe((confirmed) => {
+        if (confirmed) this.store.dispatch(new CreateWorkflow());
+      });
+  }
+
+  updateWorkflow() {
```
kevinwallimann, line 134: I wouldn't have a confirmation dialog for updating the workflow. It kind of discourages saving

ui/src/app/components/workflows/workflow/workflow.component.ts
```diff
@@ -121,6 +123,22 @@ export class WorkflowComponent implements OnInit, OnDestroy {
       });
   }
 
+  createWorkflow() {
+    this.confirmationDialogServiceSubscription = this.confirmationDialogService
```
kevinwallimann, line 127: Also for create, I would omit the confirmation dialog. If someone accidentally creates a workflow, the user can delete it afterwards

ui/src/app/stores/workflows/workflows.reducers.ts
```diff
@@ -320,6 +320,58 @@ export function workflowsReducer(state: State = initialState, action: WorkflowsA
...
+      return {
+        ...state,
+        projects: [...projects],
+        loading: false,
+      };
+    case WorkflowsActions.CREATE_WORKFLOW_FAILURE:
+      return {
+        ...state,
+        loading: false,
+      };
```
kevinwallimann, line 355: If there is an error on create workflow, all the inputs are cleared, which should not be the case. That's the case because for some reason, route.params receives a change and since we're subscribed to it in `workflow.component.ts`, `StartWorkflowInitialization` is triggered, which clears all the inputs.

It works for the `UPDATE_WORKFLOW_FAILURE` case

ui/src/app/stores/workflows/workflows.reducers.ts
```diff
@@ -320,6 +320,58 @@ export function workflowsReducer(state: State = initialState, action: WorkflowsA
...
+    case WorkflowsActions.UPDATE_WORKFLOW:
+      return {
+        ...state,
+        loading: true,
+      };
+    case WorkflowsActions.UPDATE_WORKFLOW_SUCCESS: {
+      const projectsWithoutWorkflow = state.projects.map((project) => {
+        return { name: project.name, workflows: project.workflows.filter((workflow) => workflow.id != action.payload.id) };
```
kevinwallimann: I would use `new ProjectModel` here, since you're using it below as well

ui/src/app/models/workflowRequest.model.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  private createWorkflowRequestObject(): Record<string, any> {
+    const workflowRequestObject = {};
+    this.detailsData.forEach((detail) => {
+      set(workflowRequestObject, detail.property, detail.value);
+    });
+
+    this.sensorData.forEach((sensor) => {
+      set(workflowRequestObject, 'sensor.' + sensor.property, sensor.value);
```
kevinwallimann: This implementation is pretty shaky imo. If anything changes in the `WorkflowJoinedModel` it has to be reflected here as well. The connection between `WorkflowRequestModel` and `WorkflowJoinedModel` is not evident at all from the code.
I would have expected a mapping from `WorkflowDataModel` to `WorkflowJoinedModel`, in some sense the reverse of the implementation in `WorkflowDataModel`

ui/src/app/components/workflows/workflow/workflow.component.ts
jozefbakus: I would live it. I kinda like it. For me it is guard for accidental creates/updates. Don't ask how, but I sometimes accidentally click on save button.

ui/src/app/components/workflows/workflow/workflow.component.ts
jozefbakus: I would live it. I kinda like it. For me it is guard for accidental creates/updates. Don't ask how, but I sometimes accidentally click on save button.

ui/src/app/components/workflows/workflow/workflow.component.ts
kevinwallimann: Ok, I have a different opinion. Let's ask Sasa about this in the future

ui/src/app/stores/workflows/workflows.reducers.ts
jozefbakus: Fixed.

ui/src/app/stores/workflows/workflows.reducers.ts
jozefbakus: Fixed.

ui/src/app/models/workflowRequest.model.ts
jozefbakus: Fixed

ui/src/app/models/workflowRequest.model.ts
```diff
@@ -16,37 +16,103 @@
...
+    public id?: number,
+    public updated?: Date,
+  ) {}
+}
+
+class WorkflowSensor {
+  constructor(public workflowId?: number, public sensorType?: { name: string }, public properties?: PropertiesModel, public id?: number) {}
+}
+
+class WorkflowJobDefinition {
```
kevinwallimann, line 39: Why did you duplicate this from `jobDefinition.model.ts`?

kevinwallimann (commented):
I'm somewhat unhappy about the naming of some of our classes.
We have `WorkflowRequest` and `WorkflowData` which are in reality just conversions from the frontend to the backend representation of a workflow. Their relation should be reflected in their names.
Also we have classes like `JobDefinition` and `JobInstance` in the same folder as `JobEntry`. I think we should put all classes related to the frontend-representation in a separate folder. As for this PR, I think it makes no sense to refactor that, but I would propose it for the next sprint.

ui/src/app/models/workflowRequest.model.ts
jozefbakus: WorkflowDetails, WorkflowSensor, and WorkflowJobDefinition are duplicates. To make mapping and use generic data, I need class copies with all optional fields.( I can create empty object and set it dynamically and do the mapping.)

jozefbakus:
> I'm somewhat unhappy about the naming of some of our classes.
> We have `WorkflowRequest` and `WorkflowData` which are in reality just conversions from the frontend to the backend representation of a workflow. Their relation should be reflected in their names.
> Also we have classes like `JobDefinition` and `JobInstance` in the same folder as `JobEntry`. I think we should put all classes related to the frontend-representation in a separate folder. As for this PR, I think it makes no sense to refactor that, but I would propose it for the next sprint.

Good idea!

ui/src/app/models/workflowRequest.model.ts
kevinwallimann: Ah I see. This can probably be remedied by using plain objects. I'll try to remove the duplicates in #168

kevinwallimann (approved):

kevinwallimann:
> > I'm somewhat unhappy about the naming of some of our classes.
> > We have `WorkflowRequest` and `WorkflowData` which are in reality just conversions from the frontend to the backend representation of a workflow. Their relation should be reflected in their names.
> > Also we have classes like `JobDefinition` and `JobInstance` in the same folder as `JobEntry`. I think we should put all classes related to the frontend-representation in a separate folder. As for this PR, I think it makes no sense to refactor that, but I would propose it for the next sprint.
> 
> Good idea!

#173

PR #174, opened 2020-05-21
Feature/168 plain objects

### LINKED ISSUE #168: Refactoring: Use plain objects in reducers
According to the redux documentation, it is "highly recommended to only put plain serializable objects" into the redux store. https://redux.js.org/faq/organizing-state#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state
Typescript classes are not serializable, because they lose their type information at serialization. Especially, methods will be lost, and also access modifiers. Currently, the typescript classes for redux in the codebase have no methods and all the fields are public so we have no problems. However, this can be enforced by using plain objects instead of typescript classes.

1. E.g. `WorkflowModel` can be rewritten like this:
```
export type WorkflowModel = {
    name: string;
    isActive: boolean;
    project: string;
    created: Date;
    updated: Date;
    id: number;
}
```
Existing usages of `WorkflowModel` don't need to be changed, except for code using the constructor (See point 3).

2. Types like this can be used in the reducers instead of explicit plain object types, e.g.
```
    workflowData: {
      details: { property: string; value: any }[];
```

3. Replacing classes with types breaks code relying on the constructor. To mitigate this, a `WorkflowModelFactory` can be added.
```
export class WorkflowModelFactory {
   static create(name: string, isActive: boolean /* ....*/): WorkflowModel {
      return {name: name; isActive: isActive /*...*/};
   }
}
```
Occurrences of `new WorkflowModel` have to be replaced with `WorkflowModelFactory.create`

4. Affected classes
Aside of `WorkflowModel`, all classes that are referenced in the reducers, as well as classes used in those classes, need to be rewritten as outlined in steps 1-3.

kevinwallimann:
Non-repetitive changes:
- `workflows.reducers.ts`
- `workflow.component.ts`
- `workflowRequest.model.ts`

### DIFF
--- a/src/test/scala/za/co/absa/hyperdrive/trigger/api/rest/services/WorkflowFixture.scala
+++ b/src/test/scala/za/co/absa/hyperdrive/trigger/api/rest/services/WorkflowFixture.scala
@@ -75,5 +75,4 @@ object WorkflowFixture {
       )
     )
   }
-
 }
--- a/ui/src/app/components/runs/filters/datetime-range-filter/datetime-range-filter.component.spec.ts
+++ b/ui/src/app/components/runs/filters/datetime-range-filter/datetime-range-filter.component.spec.ts
@@ -16,7 +16,7 @@
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 
 import { DatetimeRangeFilterComponent } from './datetime-range-filter.component';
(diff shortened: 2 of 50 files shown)


jozefbakus (approved):
Code looks fine. ✓
Logically it is fine. ✓
I did regress testing. ✓
I've tried find some other solution for Factory classes. (no better solution). ✓
Nice. Approved

PR #175, opened 2020-05-22
Feature/160 create test data

kevinwallimann:
**How to use**
1. Create a local database on port 5432 named "hyperdriver" and a user also named "hyperdriver" (See `LocalDatabaseConfig` for config)
2. Execute initial sql-script `db_script_v1.sql`
3. Execute test `TestDataGeneratorLocal` (in IntelliJ). If given the choice, execute with scalatest, not Junit.
4. Play around
5. Truncate all tables like this `truncate event, job_instance, job_definition, dag_instance, dag_definition, sensor, workflow;`
6. Repeat from step 3

Let me know if the handling is acceptable, or you wish to execute from cmdline, nicer random strings etc.


src/test/scala/za/co/absa/hyperdrive/trigger/testdata/LocalDatabaseConfig.scala
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package za.co.absa.hyperdrive.trigger.testdata
+
+import org.scalatest.{BeforeAndAfterAll, Suite}
+
+trait LocalDatabaseConfig extends BeforeAndAfterAll { this: Suite =>
+
+  override def beforeAll(): Unit = {
+    System.setProperty("db.driver", "org.postgresql.Driver")
```
jozefbakus, line 24: Is it possible to use application properties file instead of this ? (I mean, without this class).

src/test/scala/za/co/absa/hyperdrive/trigger/TestUtils.scala
```diff
@@ -0,0 +1,28 @@
...
+
+package za.co.absa.hyperdrive.trigger
+
+import java.util.concurrent.TimeUnit
+
+import scala.concurrent.{Await, Future}
+import scala.concurrent.duration.Duration
+
+object TestUtils {
+  def await[T](future: Future[T]): T = {
```
jozefbakus, line 25: It would be fine to replace all occurrences of this in tests. We can do it later.

jozefbakus (commented):
Random strings are fine. But I execute it via cmdline would be perfect.
+ Would be fine to have different sensors and job types data.

src/test/scala/za/co/absa/hyperdrive/trigger/testdata/LocalDatabaseConfig.scala
kevinwallimann: Not entirely. From the spring side, I can use `@TestPropertySource` to define another properties file to use than the default one. However, slick obviously doesn't know about this. By default, Slick loads any configuration file with name `application.conf`, `application.json` or `application.properties` which is on the classpath. (See `JdbcBackend.forConfig`, `DefaultConfigLoadingStrategy.parseApplicationConfig`, `SimpleIncluder.fromBasename`). 

An alternative properties file can be specified with the system properties `config.resource`, `config.file` or `config.url` (which are unfortunately very general names, could easily be conflicted imo). So I could extract the three properties to a separate file, but I would still have to define the above mentioned system property.

src/test/scala/za/co/absa/hyperdrive/trigger/TestUtils.scala
kevinwallimann: Yes I didn't want to touch all those tests in this PR

jozefbakus: Ok

kevinwallimann:
> Random strings are fine. But I execute it via cmdline would be perfect.
> 
> * Would be fine to have different sensors and job types data.

Updated PR
You can do now: `mvn test -Ptestdata` to create the testdata. Also added different sensor and job type

jozefbakus (approved):
👍🏻

PR #181, opened 2020-05-29
Generate quartz job id on insert and update

kevinwallimann:


src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/WorkflowService.scala
```diff
@@ -78,30 +82,31 @@ class WorkflowServiceImpl(override val workflowRepository: WorkflowRepository,
   }
 
   override def updateWorkflow(workflow: WorkflowJoined)(implicit ec: ExecutionContext): Future[Either[Seq[ApiError], WorkflowJoined]] = {
+    val workflowWithQuartzJobId = addQuartzJobIdForTimeSensor(workflow)
```
jozefbakus, line 85: Do we really need uiid ? Isn't enough to use sensor id. Quartz job key consists from name + group, which should be unique enough (if not, we can have uiid in group.) This way we can have service methods generic for all sensor types.

kevinwallimann: You're absolutely right. The quartzJobId never changes and has a 1-to-1 relationship with the sensorId. That means I can get rid of the quartzJobId.

PR #186, opened 2020-06-04
Use sensorId for quartz jobId instead of uuid

### LINKED ISSUE #178: Generate quartzJobId
For the time sensor, the variable "quartzJobId" is required. In the old UI, it was generated in https://github.com/AbsaOSS/hyperdrive-trigger/blob/develop/src/main/resources/ui/controller/sensor/TimeController.js 

Since the "quartzJobId" is a generated value, it should be set in the backend upon creation / update. Note: Make sure that the whole workflow is returned with the generated value.

kevinwallimann:
Closes #178


jozefbakus (approved):
OK

PR #207, opened 2020-06-17
init impl

### LINKED ISSUE #192: Update delete workflow - side nav is not updated correctly
1. Delete workflow - When Project contains only one workflow and that workflow is deleted, it shows empty project in side navigation. - Project should be removed too.

2. Update workflow - When Project contains only one workflow and that workflow is updated with new project name, it shows old project name in side navigation. - Project should be updated too.

senelesithole:
fix #192

### DIFF
--- a/ui/src/app/stores/workflows/workflows.reducers.ts
+++ b/ui/src/app/stores/workflows/workflows.reducers.ts
@@ -360,7 +360,6 @@ export function workflowsReducer(state: State = initialState, action: WorkflowsA
       };
     case WorkflowsActions.UPDATE_WORKFLOW_SUCCESS: {
       const projectsWithoutWorkflow = state.projects.map((project) => {
-        // if (project.workflows.length > 1) {
         return ProjectModelFactory.create(
           project.name,
           project.workflows.filter((workflow) => workflow.id != action.payload.id),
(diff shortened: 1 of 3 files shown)


kevinwallimann (changes requested):
Imo, this doesn't address the root cause of the bug. The problem is probably in the Workflow Reducer, e.g. in the case `WorkflowsActions.DELETE_WORKFLOW_SUCCESS` you can see that old projects are never deleted from the store, unless the whole page is refreshed ofc.
You can see in the screenshot (from redux dev tools) that the rename of the project simply lead to an addition of a new project instance in the store.
image
I guess the underlying confusion is because in the data model, `workflowModel.project` is thought to be merely a reference to project, while the definition of the project name is in `ProjectModel.name`. However in the UI, the Project Name input field simultaneously serves as both reference and definition of a project name. Honestly I think we haven't done ourselves a favour to not cleanly separate project definition from usage, but now we have to live with it.

ui/src/app/stores/workflows/workflows.reducers.ts
```diff
@@ -245,12 +245,13 @@ export function workflowsReducer(state: State = initialState, action: WorkflowsA
         loading: true,
```
jozefbakus, line 245: Cool

jozefbakus (approved):

kevinwallimann (approved):
That was easier than I expected 👌

jozefbakus:
> That was easier than I expected 👌

I thought same thing :D

senelesithole:
The lines of code don't really reflect the complexity of the issue 😊.

so that wasn't easy.

Thanks, Guys.

PR #222, opened 2020-06-23
run detail order/sort by finished column

### LINKED ISSUE #212: UI - Run detail - incorrect order of jobs
In RunDetail view, the order of jobs is not correct. ( should be order by order ) 
image

senelesithole:
fix #212


kevinwallimann:
According to the issue, the jobs should be ordered by job order. If you sort it by started or finished time, the sorting will be undefined /random when the jobs haven't started yet.

jozefbakus (changes requested):
I would solve this on effect side.
Response for job instances contains order field, but it is not defined in js class. 
So you have to add order: number into JobInstanceModel in jobInstance.model.ts.
And update runDetailGet in runs.effects.ts (sort instances that go into store somewhere here: 
`return [
            {
              type: RunActions.GET_DAG_RUN_DETAIL_SUCCESS,
              payload: jobInstance,
            },
          ];`)

ui/src/app/stores/runs/runs.effects.ts
```diff
@@ -59,7 +59,7 @@ export class RunsEffects {
           return [
             {
               type: RunActions.GET_DAG_RUN_DETAIL_SUCCESS,
-              payload: jobInstance,
+              payload: jobInstance.sort((jobInstanceA, jobInstanceB) => jobInstanceA.order - jobInstanceB.order),
```
jozefbakus, line 62: Could you please write test for this piece.

ui/src/app/stores/runs/runs.effects.spec.ts
```diff
@@ -90,24 +90,48 @@ describe('RunsEffects', () => {
...
+          new Date(Date.now()),
+          JobStatusFactory.create('Status'),
+          3,
         ),
       ];
 
       const action = new GetDagRunDetail(id);
       mockActions = cold('-a', { a: action });
-      const dagRunDetailResponse = cold('-a|', { a: jobInstances });
+      const dagRunDetailResponse = cold('-a|', { a: [jobInstancesA, jobInstancesB, jobInstancesC] });
```
jozefbakus: Could you please change this to CBA ?

jozefbakus (approved):

PR #225, opened 2020-06-26
impl: serialization key-string-value

### LINKED ISSUE #210: UI - kafka sensor - match property incorrect serialization
In the Kafka sensor, when adding a match property with a dot in the name, it's serialized as an object not as a string, thus can not be saved. 
Should be serialized as `matchProperties: {"some.key":"some.value"}`
First noticed in 01c4175f54056702d68206ea6e7ddba2eecf3c9b
image

image

senelesithole:
fixes #210


jozefbakus (approved):
👍🏻

PR #242, opened 2020-07-08
Upgrade/219 upgrade clarity

### LINKED ISSUE #219: Clarity upgrade to 3.1.4
Clarity upgrade to 3.1.4
How to do it : https://clarity.design/documentation/get-started

`ng update @clr/angular`

senelesithole:
resolves #219


ui/tslint.json
```diff
@@ -1,8 +1,16 @@
 {
```
kevinwallimann, line 1: Interesting stuff. Was it added automatically? I guess the `tslint.json` doesn't affect us anyway since we're using eslint

kevinwallimann (commented):
There are a few warnings like "depends on 'lodash/cloneDeep'. CommonJS or AMD dependencies can cause optimization bailouts."
According to https://github.com/angular/angular-cli/issues/18020, 
https://stackoverflow.com/questions/62589229/angular-10-upgrade-fix-commonjs-or-amd-dependencies-can-cause-optimization-bai/62589268#62589268, this can be remedied by replacing `lodash` with `lodash-es`

ui/tslint.json
senelesithole: Yes, it was added automatically. This happens during the 'migrations of packages' process.

senelesithole:
I actually saw those links but my worry was if lodash-es was providing the same functionality as lodash.

kevinwallimann:
Should've commented that earlier, but I don't see a reason why a minor upgrade of clarity should require a major upgrade of angular. I guess it's a bit unfortunate that clarity recommends upgrading angular in their doc, only to log a warning that it requires angular 9 and not 10. I think we should probably wait for clarity and ngrx until they officially support angular 10. So I would revert everything and only upgrade clarity. @jozefbakus What do you think?

jozefbakus:
> Should've commented that earlier, but I don't see a reason why a minor upgrade of clarity should require a major upgrade of angular. I guess it's a bit unfortunate that clarity recommends upgrading angular in their doc, only to log a warning that it requires angular 9 and not 10. I think we should probably wait for clarity and ngrx until they officially support angular 10. So I would revert everything and only upgrade clarity. @jozefbakus What do you think?

jozefbakus:
> Should've commented that earlier, but I don't see a reason why a minor upgrade of clarity should require a major upgrade of angular. I guess it's a bit unfortunate that clarity recommends upgrading angular in their doc, only to log a warning that it requires angular 9 and not 10. I think we should probably wait for clarity and ngrx until they officially support angular 10. So I would revert everything and only upgrade clarity. @jozefbakus What do you think?

Yes, I agree. @senelesithole can you revert all changes and execute only `ng update @clr/angular` ?

jozefbakus:
Sorry, I accidentally closed your PR

PR #247, opened 2020-07-14
fixed leading amd trailing whitespaces

### LINKED ISSUE #237: Whitespace in user-input fields should be trimmed before saving
@kevinwallimann 
Whitespace in user-input fields should be trimmed before saving. In my case, I had a whitespace at the beginning of the driver jar path. As a result, the job couldn't be submitted (no error message, status just went to "Lost"). I've only seen the whitespace when I looked at it in the DB.

Screenshot 2020-07-08 at 10 32 10

senelesithole:
fixes #237


kevinwallimann (approved):
lgtm

jozefbakus (approved):
Nice

PR #248, opened 2020-07-16
Feature/214 runs filters extraction v2

### LINKED ISSUE #214: UI - runs filters extraction and use them in workflows datagrid
UI - runs filters should be extracted outside of runs folder that they can be used in workflows data grid as well.

Screenshot 2020-06-22 at 15 06 18

### LINKED ISSUE #215: UI - Workflows datagrid filters/sorts
UI - Workflows datagrid filters/sorts. (Should be implemented same way as it is in Runs datagrid)
+ reset buttons.

blocked by: #214 

Screenshot 2020-06-22 at 15 07 19

senelesithole:
closes #214, closes #215


ui/src/app/components/workflows/workflows-home/workflows-home.component.ts
```diff
@@ -78,9 +102,39 @@ export class WorkflowsHomeComponent implements OnInit, OnDestroy {
     this.router.navigate([absoluteRoutes.SHOW_WORKFLOW, id]);
   }
 
+  onClarityDgRefresh(state: ClrDatagridStateInterface) {
+    if (!this.ignoreRefresh) {
+      this.sort = state.sort ? new SortAttributesModel(state.sort.by as string, state.sort.reverse ? -1 : 1) : undefined;
+      this.store.dispatch(new SetWorkflowsSort(this.sort));
```
kevinwallimann, line 108: Why is it necessary to dispatch an event? What's the difference to `RunsComponent`, where it's not needed?

kevinwallimann (commented):
Functionality is ok.

ui/src/app/components/workflows/workflows-home/workflows-home.component.ts
senelesithole: Yes, it is necessary, RunsComponents fetches the data from the server, while here the idea was to filter/sort only from the front-end, so to achieve this is by dispatching/keeping the current sort and/or filter state(s) to the store.

jozefbakus: Yes, Senele is right.

kevinwallimann (approved):

jozefbakus (approved):
👍🏻

PR #252, opened 2020-07-21
order of workflows on the left menu sorted

### LINKED ISSUE #241: Order of workflows in the left menu is not preserved
@kevinwallimann 
Order of workflows in the left menu is not preserved after updating a workflow and reopening the same project in the left menu.

During loading left menu on workflows page
projects and workflows should be sorted.

 During any action on workflow,
workflows and projects should be reordered 
to achieve no changes in orders.

senelesithole:
fixes #241 

During loading left menu on workflows page projects and workflows should be sorted. :  **completed**

During any action on workflow,
workflows and projects should be reordered
to achieve no changes in orders. 
**WIP for workflows,  projects work fine**

### DIFF
--- a/ui/src/app/components/workflows/workflow/workflow.component.ts
+++ b/ui/src/app/components/workflows/workflow/workflow.component.ts
@@ -26,6 +26,7 @@ import {
   RunWorkflow,
   SwitchWorkflowActiveState,
   RemoveBackendValidationError,
+  InitializeWorkflows,
 } from '../../../stores/workflows/workflows.actions';
 import { workflowModes } from '../../../models/enums/workflowModes.constants';
 import { absoluteRoutes } from '../../../constants/routes.constants';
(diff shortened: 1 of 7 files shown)


ui/src/app/components/workflows/workflow/workflow.component.ts
```diff
@@ -159,13 +160,22 @@ export class WorkflowComponent implements OnInit, OnDestroy {
...
+          }
         });
     } else {
       this.showHiddenParts();
     }
   }
 
+  initializeAfterAction() {
+    setTimeout(() => {
+      this.store.dispatch(new InitializeWorkflows());
```
kevinwallimann, line 175: This looks fishy to me :) Why do you need to invoke Initialize after Update? Can't you sort the projects in the `UPDATE_WORKFLOW_SUCCESS` case of the reducer?

PR #256, opened 2020-07-24
Feature/255 update sensors

### LINKED ISSUE #255: Changes in sensors are not reflected
Kevin: 
> Currently, when changing the time sensor cron expression, the change is only reflected in the quartz scheduler when the job is deactivated, wait for one scheduler heartbeat, and activate the job again. This is highly unintuitive. Either the quartz job should be restarted with the new cron expression on save, or the scheduler should check in every heartbeat if the cron expression of the time sensor has changed and restart the quartz-job.

One of the possible solutions (and discussed is to check already created sensors against values in db)

kevinwallimann:
Closes #255

### DIFF
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/persistance/SensorRepository.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/persistance/SensorRepository.scala
@@ -16,13 +16,15 @@
 package za.co.absa.hyperdrive.trigger.persistance
 
 import org.springframework.stereotype
-import za.co.absa.hyperdrive.trigger.models.Sensor
+import za.co.absa.hyperdrive.trigger.models.{Properties, Sensor}
+import za.co.absa.hyperdrive.trigger.models.enums.SensorTypes.SensorType
 
 import scala.concurrent.{ExecutionContext, Future}
 
 trait SensorRepository extends Repository {
   def getNewActiveSensors(idsToFilter: Seq[Long])(implicit ec: ExecutionContext): Future[Seq[Sensor]]
   def getInactiveSensors(ids: Seq[Long])(implicit ec: ExecutionContext): Future[Seq[Long]]
+  def getChangedSensors(originalSensors: Seq[Sensor])(implicit ec: ExecutionContext): Future[Seq[Sensor]]
 }
 
 @stereotype.Repository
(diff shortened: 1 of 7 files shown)


src/main/scala/za/co/absa/hyperdrive/trigger/persistance/SensorRepository.scala
```diff
@@ -16,13 +16,15 @@
 package za.co.absa.hyperdrive.trigger.persistance
 
 import org.springframework.stereotype
-import za.co.absa.hyperdrive.trigger.models.Sensor
+import za.co.absa.hyperdrive.trigger.models.{Properties, Sensor}
```
jozefbakus: You can remove: 
import za.co.absa.hyperdrive.trigger.models.Properties
import za.co.absa.hyperdrive.trigger.models.enums.SensorTypes.SensorType

src/main/scala/za/co/absa/hyperdrive/trigger/persistance/SensorRepository.scala
```diff
@@ -46,4 +48,23 @@ class SensorRepositoryImpl extends SensorRepository {
       sensor.id
     }).result
   }
+
+  override def getChangedSensors(originalSensors: Seq[Sensor])(implicit ec: ExecutionContext): Future[Seq[Sensor]] = db.run {(
+    for {
+      sensor <- sensorTable if originalSensors
+        .map(originalSensor => sensorIsDifferent(sensor, originalSensor))
+        .reduceLeftOption(_ || _)
```
jozefbakus, line 55: Nice

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/sensors/Sensors.scala
```diff
@@ -43,6 +43,7 @@ class Sensors @Inject()(eventProcessor: EventProcessor, sensorRepository: Sensor
   def processEvents(): Future[Unit] = {
     logger.debug(s"Processing events. Sensors: ${sensors.keys}")
     val fut = for {
+      _ <- updateChangedSensors()
```
jozefbakus: I would change order. _ <- removeInactiveSensors() and after that  _ <- updateChangedSensors()

jozefbakus (dismissed):

jozefbakus (approved):

PR #257, opened 2020-07-28
workflow history implementation

### LINKED ISSUE #233: Workflow Versioning
Keep track of workflow changes and person who did a change. 
Technical details: 
- Create history table for workflow with fields who, when, operation and old workflow in json (with db schema validation for json).
- During any operation performed on backend (Create, Update, Delete) save old workflow version into history table.
- Show history on FE.

jozefbakus:
implements: #233 

Do not forget apply db changes. 
I've skipped two tests in WorkflowsEffects.


src/main/resources/db_scripts/db_script_v3.sql
```diff
@@ -0,0 +1,149 @@
+/*
```
kevinwallimann, line 1: Why do we actually have these `db_script` files? Except for `v0` they are redundant and only add copy-paste risks

src/main/scala/za/co/absa/hyperdrive/trigger/persistance/WorkflowRepository.scala
```diff
@@ -19,43 +19,55 @@ import java.time.LocalDateTime
...
+      (for {
+        workflowId <- workflowTable returning workflowTable.map(_.id) += workflow.toWorkflow.copy(created = LocalDateTime.now())
+        sensorId <- sensorTable returning sensorTable.map(_.id) += workflow.sensor.copy(workflowId = workflowId)
+        dagId <- dagDefinitionTable returning dagDefinitionTable.map(_.id) += workflow.dagDefinitionJoined.toDag().copy(workflowId = workflowId)
+        jobId <- jobDefinitionTable returning jobDefinitionTable.map(_.id) ++= workflow.dagDefinitionJoined.jobDefinitions.map(_.copy(dagDefinitionId = dagId))
+      } yield {
+        workflowId
+      }).flatMap(id => {
+        getWorkflowJoined(id).map(
+          workflowUpdated => workflowHistoryRepository.create(workflowUpdated, user)
```
kevinwallimann, line 62: Can't you do this in the for-comprehension above? Why do you need to call `getWorkflowJoined`?

src/main/scala/za/co/absa/hyperdrive/trigger/persistance/WorkflowRepository.scala
```diff
@@ -109,43 +123,53 @@ class WorkflowRepositoryImpl extends WorkflowRepository {
...
       w <- workflowTable.filter(_.id === workflow.id).update(workflow.toWorkflow.copy(updated = Option(LocalDateTime.now())))
       s <- sensorTable.filter(_.workflowId === workflow.id).update(workflow.sensor)
       dd <- dagDefinitionTable.filter(_.workflowId === workflow.id).update(workflow.dagDefinitionJoined.toDag())
       deleteJds <- jobDefinitionTable.filter(_.dagDefinitionId === workflow.dagDefinitionJoined.id).delete
       insertJds <- jobDefinitionTable ++= workflow.dagDefinitionJoined.jobDefinitions.map(_.copy(dagDefinitionId = workflow.dagDefinitionJoined.id))
-    } yield {}).transactionally.asTry
-    ).map {
+    } yield {
+      w
+    }).asTry.map {
```
kevinwallimann: I think the `.asTry.map` section should be the last one. If there's an error while inserting the history entry, the error will be swallowed silently.

kevinwallimann: Same in `switchWorkflowActiveState`

src/main/scala/za/co/absa/hyperdrive/trigger/persistance/WorkflowHistoryRepository.scala
```diff
@@ -0,0 +1,98 @@
...
+      (for {
+        leftWorkflowHistory <- workflowHistoryTable if leftWorkflowHistory.id === leftWorkflowHistoryId
+        rightWorkflowHistory <- workflowHistoryTable if rightWorkflowHistory.id === rightWorkflowHistoryId
+      } yield {
+        (leftWorkflowHistory, rightWorkflowHistory)
+      }).result
+    )
+
+    queryResult.map(
+      _.headOption.map{ workflowsForComparisonTouple =>
```
kevinwallimann: Typo in Touple

src/test/scala/za/co/absa/hyperdrive/trigger/persistance/WorkflowHistoryRepositoryTest.scala
```diff
@@ -0,0 +1,111 @@
...
+
+  override def afterAll: Unit = {
+    h2SchemaDrop()
+  }
+
+  override def afterEach: Unit = {
+    clearData()
+  }
+
+  private def historyVerification(history: Seq[History], historyId: Long, user: String, dbOperation: DBOperation): Boolean = {
```
kevinwallimann: Method names should be verbs

src/test/scala/za/co/absa/hyperdrive/trigger/persistance/WorkflowHistoryRepositoryTest.scala
```diff
@@ -0,0 +1,111 @@
...
+
+  "workflowHistoryRepository create/update/delete/getHistoryForWorkflow" should "create workflow history record with correct values" in {
+    val workflowCreate = WorkflowFixture.createWorkflowJoined()
+    val workflowUpdate = workflowCreate.copy(name = workflowCreate.name + "-updated")
+    val workflowDelete = workflowCreate.copy(name = workflowCreate.name + "-deleted")
+
+    val workflowId = workflowCreate.id
+    val user = "fakeUser"
+
+    val createResultId = Await.result(db.run(workflowHistoryRepository.create(workflowCreate, user)), Duration.Inf)
```
kevinwallimann: You may want to use `TestUtils.await`. Potentially waiting for infinity might not be the best thing to do.

src/test/scala/za/co/absa/hyperdrive/trigger/persistance/WorkflowHistoryRepositoryTest.scala
```diff
@@ -0,0 +1,111 @@
...
+  override def afterAll: Unit = {
+    h2SchemaDrop()
+  }
+
+  override def afterEach: Unit = {
+    clearData()
+  }
+
+  private def historyVerification(history: Seq[History], historyId: Long, user: String, dbOperation: DBOperation): Boolean = {
+    history.exists(history => {
```
kevinwallimann: Maybe rename `history` to `historyEntries` ;)

src/main/scala/za/co/absa/hyperdrive/trigger/models/WorkflowHistory.scala
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+
+package za.co.absa.hyperdrive.trigger.models
+
+import java.time.LocalDateTime
+
+import za.co.absa.hyperdrive.trigger.models.enums.DBOperation.DBOperation
+
+case class History(
```
kevinwallimann, line 22: I would rename this to `HistoryEntry`. If you think of history, you think of a sequence of events. At first sight, I assumed this class would hold a collection of events, but it describes exactly one event. Naming it `History` will lead to awkward namings in loops and collections.

kevinwallimann: Same for the `WorkflowHistory`

ui/src/app/components/workflows/workflow-form/jobs/job/job.component.ts
```diff
@@ -30,67 +29,62 @@ import { delay } from 'rxjs/operators';
...
-      this.jobData = !!jobDataOption ? jobDataOption.entries : [];
-
-      const selected = this.jobData.find((value) => value.property == this.jobSwitchPart.property);
-      this.selectedJob = !!selected ? selected.value : undefined;
-    });
-
-    this.jobChangesSubscription = this.jobChanges.pipe(delay(0)).subscribe((jobChange) => {
-      if (jobChange.property == this.jobSwitchPart.property) {
-        this.store.dispatch(
+    this.jobChangesSubscription = this.jobChanges.subscribe((jobChange) => {
```
kevinwallimann, line 47: Why is the `delay(0)` not needed anymore?

ui/src/app/components/workflows/workflow-history/workflow-history.component.ts
```diff
@@ -0,0 +1,64 @@
...
+      this.loading = state.history.loading;
+      this.workflowHistory = state.history.workflowHistory;
+    });
+  }
+
+  isSelectable(inputHistory: HistoryModel): boolean {
+    const areTwoSelected = this.selected.length == 2;
+    const isAlreadySelected = this.selected.some((history: HistoryModel) => history === inputHistory);
+    const result = areTwoSelected && !isAlreadySelected;
+    return !result;
```
kevinwallimann: There are too many negations here. Also, directly return the expression. `result` doesn't improve readability

kevinwallimann:
General comments / questions
- Maybe we should add a version number (like v1, v2) for each workflow. Will be easier for the user to talk about a specific revision
- A history entry is always created when the user saves the workflow, even if there's no change. It depends on how often the user would do it, but it would certainly "dilute" the history and diminish its usefulness.

src/main/resources/db_scripts/db_script_v3.sql
jozefbakus: Yes you are right, but now we can see whole table structure without going over v0 and all delta scripts. Maybe it would make sense to have v0 and vLast. (copy-paste risks is still there but is lower)

src/main/scala/za/co/absa/hyperdrive/trigger/persistance/WorkflowRepository.scala
jozefbakus: Yes, I can but I wanted to have it separated from insert functionality ( and potentially we can extract it into function and call only that one). Yes, we need exact copy which is updated/created (on input you already have out of date workflow, because  of .copy(created = LocalDateTime.now()). Even if I update workflow manually I will always have workflow which in db).

ui/src/app/components/workflows/workflow-form/jobs/job/job.component.ts
jozefbakus: Because it does not update component props multiple times during one component life cycle.

src/main/scala/za/co/absa/hyperdrive/trigger/persistance/WorkflowRepository.scala
jozefbakus: Fixed

src/main/scala/za/co/absa/hyperdrive/trigger/persistance/WorkflowHistoryRepository.scala
jozefbakus: Fixed

src/test/scala/za/co/absa/hyperdrive/trigger/persistance/WorkflowHistoryRepositoryTest.scala
jozefbakus: Fixed

src/test/scala/za/co/absa/hyperdrive/trigger/persistance/WorkflowHistoryRepositoryTest.scala
jozefbakus: Fixed

src/test/scala/za/co/absa/hyperdrive/trigger/persistance/WorkflowHistoryRepositoryTest.scala
jozefbakus: Fixed

ui/src/app/components/workflows/workflow-history/workflow-history.component.ts
jozefbakus: Ooooooooo boy :D. Take a look now, should be readable and straight forward.

jozefbakus:
> General comments / questions
> 
> * Maybe we should add a version number (like v1, v2) for each workflow. Will be easier for the user to talk about a specific revision
> * A history entry is always created when the user saves the workflow, even if there's no change. It depends on how often the user would do it, but it would certainly "dilute" the history and diminish its usefulness.

* It is not necessary now as discussed
* Aware of that, it is old bug (https://github.com/AbsaOSS/hyperdrive-trigger/issues/264). We will take it into next sprint

src/main/scala/za/co/absa/hyperdrive/trigger/models/WorkflowHistory.scala
jozefbakus: As discussed, I will not fix it.

src/main/resources/db_scripts/db_script_v3.sql
kevinwallimann: Ah that's the reason. I think it's better to look at the DB directly than relying on these scripts because any typo, copy-paste error will go unnoticed. But ok, we can discuss about it in the context of the liquibase feature.

jozefbakus: ok

kevinwallimann (dismissed):

kevinwallimann (approved):

PR #259, opened 2020-07-29
keep workflow screen filters on runs screen

### LINKED ISSUE #21: Add Runs button to action menu of a workflow
Many times people will only want to go to their runs when they see/view what is in them to see how is the workflow behaving.

For that reason, I would add a `Runs` button to a workflow action menu that would open `URL/#/WorkflowRuns/X` without the need to go to runs, search for your WF again and open it up.

**Main point:** We would go from 5 operations to 3
5 operations: Search for workflow, See what it is, Open Runs, Search for Workflow, Open its run
3 operations: Search for workflow, See what it is, Open its runs

senelesithole:
resolves #21 

**NB**
There’s an 'error/exception' when clearing the filters from the runs screen. 
1. [HPM] Error occurred while trying to proxy request ... -- front-end
2. org.apache.catalina.connector.ClientAbortException: java.io.IOException: Broken pipe -- back-end

did a bit of research can’t seem to pinpoint the root cause.

### DIFF
--- a/ui/src/app/components/runs/runs.component.html
+++ b/ui/src/app/components/runs/runs.component.html
@@ -33,10 +33,12 @@
       </div>
     </clr-dg-action-bar>
 
-    <clr-dg-column [clrDgField]="dagRunColumns.WORKFLOW_NAME" [clrFilterValue]>
+    <clr-dg-column [clrDgField]="dagRunColumns.WORKFLOW_NAME">
       <ng-container *clrDgHideableColumn="{hidden: false}">Workflow name</ng-container>
       <clr-dg-filter [clrDgFilter]="workflowNameFilter">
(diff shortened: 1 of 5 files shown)


kevinwallimann:
For me the feature works, I can't reproduce your error message. Looks like the backend wasn't reachable?

Anyway, maybe there was a misunderstanding in the feature request, but I think it's not exactly what we were looking for.
The description in #21 says, that a button should be added. So you should be able to go from http://localhost:4200/#/workflows/show/154 to the runs page, not from http://localhost:4200/#/workflows. Syncing the filter across the workflows page and the runs page creates is a very hidden feature and probably creates more problems than it solves.

senelesithole:
rework needed.

PR #261, opened 2020-07-30
Feature/21 add view run button

### LINKED ISSUE #21: Add Runs button to action menu of a workflow

senelesithole:
resolves #21

### DIFF
--- a/ui/src/app/components/runs/runs.component.html
+++ b/ui/src/app/components/runs/runs.component.html
@@ -33,10 +33,12 @@
       </div>
     </clr-dg-action-bar>
 
-    <clr-dg-column [clrDgField]="dagRunColumns.WORKFLOW_NAME" [clrFilterValue]>
+    <clr-dg-column [clrDgField]="dagRunColumns.WORKFLOW_NAME">
       <ng-container *clrDgHideableColumn="{hidden: false}">Workflow name</ng-container>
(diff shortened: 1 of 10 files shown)


kevinwallimann:
Turns out it's not that straight-forward. Consider the following example: I have two workflows, "Some Workflow" and "Some Workflow 2". If I click on View Runs for "Some Workflow", the runs for "Some Workflow 2" are also displayed, because the filter is a contains filter.

image

We could make the workflow name filter match only exactly, or we could add a workflow id column, or maybe we could filter for workflow id, but not show it to the user.

senelesithole:
I would leave the filter as contains. A user may want to use the filter on a long length workflow name on the runs screen, now on an exact match, the user will be required to write/copy&paste the whole long workflow name while on contains the user may need to just write a keyword. w.r.t workflow id maybe that may work.

PR #262, opened 2020-07-30
Add Git hash to manifest to identify jar

kevinwallimann:
Since we're continuing to work on snapshot versions, I think it's handy to write the git-hash into the jar, so we exactly know what the jar contains.


senelesithole (approved):

jozefbakus (approved):
👍🏻

PR #265, opened 2020-07-31
Feature/21 redirect to runs from workflow pages

jozefbakus:
Implements: #21


src/main/resources/db_scripts/db_script_v3.sql
```diff
@@ -0,0 +1,141 @@
+/*
```
kevinwallimann: Careful with the version number. I would already increase it to v4.

jozefbakus: Yes, I know. It will be showed as a conflict

jozefbakus: Solved

kevinwallimann (approved):

PR #269, opened 2020-08-03
datetime-range-filter max min date null error

### LINKED ISSUE #266: Runs datagrid - setMaxDate/setMinDate null error
datetime range filter html component ([max]="value.to | date:'yyyy-MM-dd'" and [min]="value.from | date:'yyyy-MM-dd'") shows null error when value.to/value.from is null/undefined.

Possible solution, we can create Min and Max date and use these values when value.from or value.to is undefined.

Screenshot 2020-08-03 at 08 27 59

jozefbakus:
Implements: #266 
At the end issue was with DateIOService which was null, because no form was initialised.


senelesithole (approved):
that was a quick one 👍🏽

kevinwallimann (approved):
Very cool. In prod mode, the UI was actually broken :)

PR #273, opened 2020-08-14
Fix/264 unnecessary workflow updates

### LINKED ISSUE #264: Unnecessary workflow updates when there is no change in workflow
When user call update workflow with values that are already persisted
db update should not be called

Two things to fix:
- [ ] FE - user should not be able to click on update button when there is no change in workflow form 
Screenshot 2020-08-03 at 12 22 42

- [ ] BE - update workflow endpoint should throw error with message saying "nothing to update" when not changed workflow is submitted. 
**PUT /workflows**

senelesithole:
fixes #264 

updateWorkflow effects tests guidance needed.

### DIFF
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/WorkflowValidationService.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/WorkflowValidationService.scala
@@ -46,7 +46,8 @@ class WorkflowValidationServiceImpl @Inject()(override val workflowRepository: W
   override def validateOnUpdate(workflow: WorkflowJoined)(implicit ec: ExecutionContext): Future[Seq[ApiError]] = {
     val validators = Seq(
       validateWorkflowIsUnique(workflow),
-      validateProjectIsNotEmpty(workflow)
+      validateProjectIsNotEmpty(workflow),
+      validateWorkflowData(workflow)
     )
     combine(validators)
   }
@@ -82,4 +83,14 @@ class WorkflowValidationServiceImpl @Inject()(override val workflowRepository: W
     }
     Future.successful(projectValidation.toSeq)
   }
+
+  private def validateWorkflowData(workflow: WorkflowJoined)(implicit ec: ExecutionContext): Future[Seq[ApiError]] = {
+    workflowRepository.getWorkflow(workflow.id)
+      .map(persistedWorkflowData => if (persistedWorkflowData == workflow) {
+        Seq(ValidationError("nothing to update"))
+      } else {
+        Seq()
+      })
+  }
 }
+
--- a/ui/src/app/components/workflows/workflow-form/workflow-form.component.html
+++ b/ui/src/app/components/workflows/workflow-form/workflow-form.component.html
@@ -50,7 +50,7 @@
       <clr-icon shape="check"></clr-icon>
       Create new workflow
     </button>
-    <button type="submit" class="btn btn-sm btn-secondary" *ngIf="mode == workflowModes.EDIT">
+    <button type="submit" class="btn btn-sm btn-secondary" [disabled]="workflowForm.pristine" *ngIf="mode == workflowModes.EDIT">
       <clr-icon shape="check"></clr-icon>
       Update workflow
     </button>
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -61,9 +61,9 @@ sparkYarnSink.additionalConfs.spark.executor.memory=2g
 
 #Postgresql properties for connection to trigger metastore
 db.driver=org.postgresql.Driver
-db.url=jdbc:postgresql://
-db.user=
-db.password=
+db.url=jdbc:postgresql://localhost/senele
+db.user=postgres
+db.password=pass
 db.keepAliveConnection=true
 db.connectionPool=HikariCP
 db.numThreads=4
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -61,9 +61,9 @@ sparkYarnSink.additionalConfs.spark.executor.memory=2g
 
 #Postgresql properties for connection to trigger metastore
 db.driver=org.postgresql.Driver
-db.url=jdbc:postgresql://localhost/senele
-db.user=postgres
-db.password=pass
+db.url=jdbc:postgresql://
+db.user=
+db.password=
 db.keepAliveConnection=true
 db.connectionPool=HikariCP
 db.numThreads=4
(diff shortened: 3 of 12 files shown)


ui/src/app/components/workflows/workflow-form/workflow-form.component.html
```diff
@@ -50,7 +50,7 @@
       <clr-icon shape="check"></clr-icon>
       Create new workflow
     </button>
-    <button type="submit" class="btn btn-sm btn-secondary" *ngIf="mode == workflowModes.EDIT">
+    <button type="submit" class="btn btn-sm btn-secondary" [disabled]="workflowForm.pristine" *ngIf="mode == workflowModes.EDIT">
```
jozefbakus, line 53: This is cool, but it does not work. 
Example:  - go to update workflow page (workflow name = xxx)
- update button is disabled. You change workflow name to yyy. (it will enable update button)
- change workflow name back to xxx and button is still enabled. 
I think you can remove it.

jozefbakus (changes requested):
Front end review: 
I would revert all changes and I would implement it in workflow-form.component this way: 
1. Implement OnInit
2. In ngOnInit  I would copy workflow: 
`this.initialWorkflowData = this.workflowData;`
3. And I would implement comparison method: 
`  hasWorkflowChanged(): boolean {
    return !(this.areWorkflowEntriesEqual(this.workflowData.details, this.initialWorkflowData.details) &&
      this.areWorkflowEntriesEqual(this.workflowData.sensor, this.initialWorkflowData.sensor) &&
      this.areJobsEqual(this.workflowData.jobs, this.initialWorkflowData.jobs));
  }`

` areWorkflowEntriesEqual(leftEntries: WorkflowEntryModel[], rightEntries: WorkflowEntryModel[]): boolean {
    return leftEntries.map((leftEntry) =>
      rightEntries.some(
        (rightEntry) =>
          isEqual(rightEntry.value, leftEntry.value) && rightEntry.property == leftEntry.property
      )
    ).every((entry) => entry);
  }`

`  areJobsEqual(leftJobEntry: JobEntryModel[], rightJobEntry: JobEntryModel[]): boolean {
    if(leftJobEntry.length != rightJobEntry.length) {
      return false;
    } else {
      const leftJobEntrySorted = leftJobEntry.slice().sort((left,right) => left.order - right.order);
      const rightJobEntrySorted = rightJobEntry.slice().sort((left,right) => left.order - right.order);
      return leftJobEntrySorted.map(
        (job, index) =>
          this.areWorkflowEntriesEqual(job.entries, rightJobEntrySorted[index].entries)
      ).every((entry) => entry);
    }
  }
`
4. Last step is to use it in html [disabled]="!hasWorkflowChanged()"

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/WorkflowValidationService.scala
```diff
@@ -82,4 +83,14 @@ class WorkflowValidationServiceImpl @Inject()(override val workflowRepository: W
     }
     Future.successful(projectValidation.toSeq)
   }
+
+  private def validateWorkflowData(workflow: WorkflowJoined)(implicit ec: ExecutionContext): Future[Seq[ApiError]] = {
+    workflowRepository.getWorkflow(workflow.id)
+      .map(persistedWorkflowData => if (persistedWorkflowData == workflow) {
```
jozefbakus, line 89: Unfortunately it will not work. Check updateWorkflow in WorkflowService. Workflow that you receive contains workflow without ids

senelesithole: saw this one hence on effects I tried to grab all the ids. thanks.

ui/src/app/components/workflows/workflow-form/workflow-form.component.html
senelesithole: noted this. thanks.

- the idea around this was for the back-end call to do the validation

jozefbakus (changes requested):
Back end review: 
Nice try, but we have to go through all workflow variables explicitly because of ids and foreign keys.
- in WorkflowValidationService.scala I would change method `validateOnUpdate` to method with different input ` def validateOnUpdate(originalWorkflow: WorkflowJoined, updatedWorkflow: WorkflowJoined)(implicit ec: ExecutionContext): Future[Seq[ApiError]]
`
- and I would change validateWorkflowDate to something like this: `private def validateWorkflowData(originalWorkflow: WorkflowJoined, updatedWorkflow: WorkflowJoined)(implicit ec: ExecutionContext): Future[Seq[ApiError]] = {
    val workflowDetailsVerification = Seq(
      originalWorkflow.name == updatedWorkflow.name,
      originalWorkflow.isActive == updatedWorkflow.isActive
    // +Others
    )

    val workflowSensorVerification = Seq(
      originalWorkflow.sensor.sensorType == updatedWorkflow.sensor.sensorType,
      originalWorkflow.sensor.properties == updatedWorkflow.sensor.properties
      // +Others
    )

    val workflowJobsVerification = Seq(
      Seq(originalWorkflow.dagDefinitionJoined.jobDefinitions.length == updatedWorkflow.dagDefinitionJoined.jobDefinitions.length),
      originalWorkflow.dagDefinitionJoined.jobDefinitions.flatMap(originalJob => {
        val updatedJobOption = updatedWorkflow.dagDefinitionJoined.jobDefinitions.find(_.order == originalJob.order)
        updatedJobOption.map(updatedJob =>
          Seq(
            originalJob.name == updatedJob.name,
            originalJob.jobType == updatedJob.jobType
            // +Others
          )
        ).getOrElse(Seq(false))
      })
    ).flatten

    if((workflowDetailsVerification ++ workflowSensorVerification ++ workflowJobsVerification).contains(false)) {
      Future.successful(Seq(ValidationError("Nothing to update")))
    } else {
      Future.successful(Seq())
    }
  }`

- and as a last step I would update updateWorkflow method in WorkflowService (this way we will query workflow only once):
`override def updateWorkflow(workflow: WorkflowJoined)(implicit ec: ExecutionContext): Future[Either[Seq[ApiError], WorkflowJoined]] = {
    val userName = getUserName.apply();
    for {
      originalWorkflow <- getWorkflow(workflow.id)
      validationErrors <- workflowValidationService.validateOnUpdate(originalWorkflow, workflow)
      result <- doIf(validationErrors, () => {
          val updatedWorkflow = workflow.copy(
            id = originalWorkflow.id,
            created = originalWorkflow.created,
            updated = originalWorkflow.updated,
            sensor = workflow.sensor.copy(
              id = originalWorkflow.sensor.id,
              workflowId = originalWorkflow.id,
              properties = workflow.sensor.properties.copy(
                sensorId = originalWorkflow.sensor.properties.sensorId
              )
            ),
            dagDefinitionJoined = workflow.dagDefinitionJoined.copy(
              id = originalWorkflow.dagDefinitionJoined.id,
              workflowId = originalWorkflow.id
            )
          )
          workflowRepository.updateWorkflow(updatedWorkflow, userName).flatMap {
            case Left(error) => Future.successful(Left(error))
            case Right(_) => getWorkflow(workflow.id).map(Right(_))
          }
      })
    } yield {
      result
    }
  }`

Issue #236, opened 2020-07-08
Simplify ingestion jobs input


jozefbakus:
We should decide how to implement simplification of ingestion job creation.

kevinwallimann:
- Add new entity `JobTemplate`
- Add new field `job_template` to `JobDefinition`

![image](https://user-images.githubusercontent.com/54136196/89418426-530d3d00-d730-11ea-965a-935145098192.png)

- `JobTemplate`s are resolved in `Executors.scala`, when `JobDefinition`s are mapped to `JobInstance`s
- To select job templates in the UI, a new rest endpoint is required to get available job templates. The selection of jobTypes will be replaced with the selection of job templates.

**Data Migration**
- Job Templates will be hard-coded to the DB for now. Create job templates for GenericSparkJob, GenericShellJob, HyperconfomranceJob, KafkaOffloadingJob
- Existing job definitions with job types Spark must be assigned the GenericSparkJob template, existing Shell jobs  must be assigend the GenericShellJob

PR #275, opened 2020-08-14
Add hyperconf template

kevinwallimann:
Linked to #236 
Only for demonstration of the next step.

You would need to create a template like this:
```
insert into "job_template" ("name", "job_type", "variables", "maps", "key_value_pairs")
values ('Hyperconformance Job', 'Spark',
'{"jobJar":"driver.jar","mainClass":"za.co.absa.hyperdrive.driver.drivers.CommandLineIngestionDriver","deploymentMode":"cluster"}',
'{"additionalJars":["spark-jobs-2.5.1-SNAPSHOT.jar"],"appArguments":["component.decoder=za.co.absa.hyperdrive.ingestor.implementation.decoder.avro.confluent.ConfluentAvroKafkaStreamDecoder","component.ingestor=Spark","component.manager=za.co.absa.hyperdrive.ingestor.implementation.manager.checkpoint.CheckpointOffsetManager","component.reader=za.co.absa.hyperdrive.ingestor.implementation.reader.kafka.KafkaStreamReader","component.transformer=za.co.absa.enceladus.conformance.HyperConformance","component.writer=za.co.absa.hyperdrive.ingestor.implementation.writer.kafka.KafkaStreamWriter"}',
'{}');
```

I believe, the predefined app arguments should not be part of the template, but be grouped as variables. E.g. there is a variable for kafka-reader and it contains component.reader, and the reader config. That way, we could also achieve one level of hierarchy. Templates could consist of variables.

### DIFF
--- a/src/main/resources/db_scripts/db_script_v5.sql
+++ b/src/main/resources/db_scripts/db_script_v5.sql
@@ -157,7 +157,7 @@ left join workflow
     on workflow.id = dag_instance.workflow_id;
 
 create table "job_template" (
-  "name" VARCHAR NOT NULL,
+  "name" VARCHAR NOT NULL UNIQUE,
   "job_type" VARCHAR NOT NULL,
   "variables" VARCHAR NOT NULL,
   "maps" VARCHAR NOT NULL,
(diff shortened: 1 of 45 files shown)


jozefbakus (commented):
WorkflowService tests fail.
Looks really good.

PR #293, opened 2020-09-03
289 - record who triggered workflow

### LINKED ISSUE #289: Record user who manually triggered workflow
Information about who manually triggered workflow should be recorded in db. 

- new field on dag_instance table (triggered_by or something similar )
- in case of automatic trigger, special value should be used ( Trigger type )
! UI impl is not required in this story

jozefbakus:
Implements: #289


kevinwallimann (approved):

kevinwallimann:
Nice!

PR #294, opened 2020-09-04
Feature/240 session timeout relogin

### LINKED ISSUE #240: Prevent data loss on session timeout
@zakera786  
While working with creating workflows , the session times out however i am only aware of this once I try to save the workflow. When saving workflow the screen re-directs me to login page. I will try and time myself so i can give correct feedback on this.
UI timed out within 10 minutes.

When user is still log in but session is expired 
popup with log-in screen should be displayed to user.
(This way all changes will not disappear)

kevinwallimann:
Closes #240 

**Notes**
- For testing, instead of waiting for session timeout, cookies can be deleted to provoke a 401


jozefbakus (approved):
Nice

PR #297, opened 2020-09-08
Feature/286 sort filter on active field

### LINKED ISSUE #286: Workflows page - sort/filter on is active field
Workflows page - sort/filter on is active field 

- Can be sort/filter reused from runs page ?
- If no implement it for this field

Screenshot 2020-09-01 at 15 27 54

senelesithole:
implements #286 
new filter: boolean filter.

**not yet working for "No" -> false boolean
reason: workflow-home component discards boolean false value.**


ui/src/app/components/runs/runs.component.html
```diff
@@ -122,7 +122,7 @@
         </clr-icon>
         <clr-icon *ngSwitchDefault
                   shape="help"
-                  style="color: darkgoldenrod"
+                  style="color: rgb(66, 63, 54)"
```
jozefbakus: Why ?

senelesithole: no reason

ui/src/app/components/common/datagrid/filters/boolean-filter/boolean-filter.component.scss
```diff
@@ -0,0 +1,9 @@
+.boolean-selectable {
```
kevinwallimann, line 16: Missing copyright

ui/src/app/components/common/datagrid/filters/boolean-filter/boolean-filter.component.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    if (this.isTrueSelected && testedValue == true) {
+      return true;
+    }
+
+    if (this.isFalseSelected && testedValue == false) {
+      return true;
+    }
+
+    return false;
```
kevinwallimann: I would inline the whole condition
`return this.isTrueSelected && ... || this.... &&`

ui/src/app/components/common/datagrid/filters/boolean-filter/boolean-filter.component.ts
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  get state() {
+    const values: string[] = [];
+    return new EqualsMultipleFilterAttributes(this.property, values);
```
kevinwallimann: Please add a comment that this is only a placeholder implementation to avoid runtime errors that should do nothing. But backend-filtering functionality is not available for this filter.

ui/src/app/components/common/datagrid/filters/boolean-filter/boolean-filter.component.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  modelChanges: Subject<any> = new Subject<any>();
+  modelSubscription: Subscription;
+
+  constructor() {
+    // do nothing
+  }
+
+  ngAfterViewInit(): void {
+    this.modelSubscription = this.modelChanges.pipe(debounceTime(500), distinctUntilChanged()).subscribe((newValue) => {
```
kevinwallimann: We should really extract the magic number 500 some time. But not now ;)

ui/src/app/components/common/datagrid/filters/boolean-filter/boolean-filter.component.ts
```diff
@@ -0,0 +1,92 @@
...
+  @Input() removeFiltersSubject: Subject<any>;
+  @Input() property: string;
+  @Input() value;
+  isTrueSelected = false;
+  isFalseSelected = false;
+
+  changes = new Subject<any>();
+
+  modelChanges: Subject<any> = new Subject<any>();
+  modelSubscription: Subscription;
```
kevinwallimann: I know we have this in the `MultipleStatusFilter` as well. But I think we don't need the indirection with the `modelChanges` subject here (and also not in the `MultipleStatusFilter`). IIRC, we only need the `modelChanges` subject for the `debounceTime` to add some delay for the filters which involve users typing. But here, we only have checkboxes, so it doesn't make sense to delay the request.

So I think, you can call `changes.next()` directly instead of `modelChanges.next()` and remove `modelSubscription` altogether.

@jozefbakus What do you think?

ui/src/app/components/common/datagrid/filters/boolean-filter/boolean-filter.component.ts
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  isActive(): boolean {
+    return this.isTrueSelected || this.isFalseSelected;
+  }
+
+  onRemoveFilter() {
+    this.isTrueSelected = false;
+    this.isFalseSelected = false;
+    this.modelChanges.next([]);
```
kevinwallimann: Value of `modelChanges` subject is actually never used. So you can just call `.next()` without arg.

ui/src/app/components/common/datagrid/filters/boolean-filter/boolean-filter.component.scss
senelesithole: fixed.

ui/src/app/components/common/datagrid/filters/boolean-filter/boolean-filter.component.ts
senelesithole: fixed.

ui/src/app/components/common/datagrid/filters/boolean-filter/boolean-filter.component.ts
senelesithole: done.

ui/src/app/components/common/datagrid/filters/boolean-filter/boolean-filter.component.ts
senelesithole: 👍🏽

ui/src/app/components/common/datagrid/filters/boolean-filter/boolean-filter.component.ts
senelesithole: fixed thanks 👍🏽 .

ui/src/app/components/common/datagrid/filters/boolean-filter/boolean-filter.component.ts
senelesithole: fixed

ui/src/app/components/common/datagrid/filters/boolean-filter/boolean-filter.component.ts
```diff
@@ -87,6 +76,6 @@ export class BooleanFilterComponent implements ClrDatagridFilterInterface<any>,
   onRemoveFilter() {
     this.isTrueSelected = false;
     this.isFalseSelected = false;
-    this.modelChanges.next([]);
+    this.modelChanges.next();
```
kevinwallimann: You need to call `changes.next()` here. `modelChanges` is not subscribed by anything anymore. `changes` is subscribed by clarity.

ui/src/app/components/common/datagrid/filters/boolean-filter/boolean-filter.component.ts
```diff
@@ -34,23 +33,18 @@ export class BooleanFilterComponent implements ClrDatagridFilterInterface<any>,
...
   constructor() {
     // do nothing
   }
 
   ngAfterViewInit(): void {
-    this.modelSubscription = this.modelChanges.pipe(debounceTime(500), distinctUntilChanged()).subscribe((newValue) => {
-      this.changes.next();
-    });
-
+    this.changes.next();
```
kevinwallimann: Do you need this here?

jozefbakus (approved):
Looks good and it works

kevinwallimann (approved):

PR #298, opened 2020-09-08
Feature/295 hyperconf template

### LINKED ISSUE #295: Add templates for hyperdrive jobs
With #236, job templates were introduced. Currently, the UI supports generic shell jobs and generic spark jobs. With this PR, UI support for hyperdrive jobs should be added. 

- On the UI, Hyperdrive jobs will only display job name, job template, additional jars, additional files, additional spark config and app arguments. 
- There can be multiple templates for various hyperdrive jobs. They will share the UI config (i.e. the same fields visible), but have different parameters in the template.

The following changes are required
- `JobTemplate` should be extended with a field `formConfig: string`
- `workflow.service.ts` should load all available job templates from the backend and render the correct form fields using `formConfig`

kevinwallimann:
Closes #295 
The sql for a sample template has been uploaded to the private "scripts" repository.

### DIFF
--- a/src/main/resources/db_scripts/db_script_v8.sql
+++ b/src/main/resources/db_scripts/db_script_v8.sql
@@ -94,7 +94,8 @@ create table "job_template" (
   "variables" VARCHAR NOT NULL,
   "maps" VARCHAR NOT NULL,
   "key_value_pairs" VARCHAR NOT NULL,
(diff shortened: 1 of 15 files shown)


jozefbakus (dismissed):
👍🏻

jozefbakus (approved):

PR #299, opened 2020-09-09
Feature/281 long running job

### LINKED ISSUE #281: Long running jobs support
Long running jobs support analysis:
- Distinguish long running jobs from normal jobs
- Respawn dead job (If workflow is active)

jozefbakus:
Closes: #281 

- new sensor type: Recurring

### DIFF
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/sensors/recurring/RecurringSensor.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/sensors/recurring/RecurringSensor.scala
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2018 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.hyperdrive.trigger.scheduler.sensors.recurring
+
+import java.time.format.DateTimeFormatter
+import java.time.Instant
+
+import org.slf4j.LoggerFactory
+import play.api.libs.json.JsObject
+import za.co.absa.hyperdrive.trigger.models.{Event, Properties, Sensor}
+import za.co.absa.hyperdrive.trigger.persistance.DagInstanceRepository
+import za.co.absa.hyperdrive.trigger.scheduler.sensors.PollSensor
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success}
+
+class RecurringSensor(
+  eventsProcessor: (Seq[Event], Properties) => Future[Boolean],
+  sensorDefinition: Sensor,
+  executionContext: ExecutionContext,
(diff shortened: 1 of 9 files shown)


src/test/scala/za/co/absa/hyperdrive/trigger/scheduler/sensors/recurring/RecurringSensorTest.scala
```diff
@@ -0,0 +1,79 @@
...
+
+    when(dagInstanceRepository.hasRunningDagInstance(eqTo(sensorDefinition.workflowId))(any[ExecutionContext])).thenReturn(Future{true})
+
+    val underTest = new RecurringSensor(eventProcessor.eventProcessor(triggeredBy), sensorDefinition, executionContext, dagInstanceRepository)
+
+    // when
+    val result: Unit = await(underTest.poll())
+
+    // then
+    result shouldBe (): Unit
```
kevinwallimann, line 60: Type annotation not required here.
Instead of checking the result (is a bit meaningless, since `result` will always be `Unit` when the code reaches this point), I would verify that `eventProcessor.eventProcessor` hasn't been called.

src/test/scala/za/co/absa/hyperdrive/trigger/scheduler/sensors/recurring/RecurringSensorTest.scala
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  "RecurringSensor.poll" should "call event processor when workflow is not running" in {
```
kevinwallimann, line 64: This test passes even if the event processor is not called

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/sensors/recurring/RecurringSensor.scala
```diff
@@ -0,0 +1,68 @@
...
+  private val eventDateFormatter: DateTimeFormatter = DateTimeFormatter.ISO_INSTANT
+
+  private val logger = LoggerFactory.getLogger(this.getClass)
+  private val logMsgPrefix = s"Sensor id = ${properties.sensorId}."
+
+  override def poll(): Future[Unit] = {
+    logger.debug(s"$logMsgPrefix. Polling new events.")
+
+    val fut = dagInstanceRepository.hasRunningDagInstance(sensorDefinition.workflowId).flatMap {
+      case hasRunningDagInstance if hasRunningDagInstance => {
```
kevinwallimann: Hmm, wouldn't if else be more readable than pattern matching here?

jozefbakus: Fixed.

src/test/scala/za/co/absa/hyperdrive/trigger/scheduler/sensors/recurring/RecurringSensorTest.scala
jozefbakus: fixed

src/test/scala/za/co/absa/hyperdrive/trigger/scheduler/sensors/recurring/RecurringSensorTest.scala
jozefbakus: 1. In case of unit, I prefer type annotation. (If you insist, I will remove it)
2. Hmm so I wrote it there, but I think it is not necessary because it is there implicitly. (Should throw exception)

kevinwallimann: Ah yes ok you are right. The same is also true for `result shouldBe()` I leave it to you

kevinwallimann (approved):

senelesithole (approved):

PR #301, opened 2020-09-09
Provide user with version and env name

### LINKED ISSUE #8: Provide user with version and env name
This request is inspired by Menas and is a quick reference to what and where are people doing.

It would be nice to see the version and Environment name/variable somewhere.

jozefbakus:
Closes: #8

### DIFF
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/UserInfoController.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/UserInfoController.scala
@@ -15,6 +15,7 @@
 
 package za.co.absa.hyperdrive.trigger.api.rest.controllers
 
+import org.springframework.beans.factory.annotation.Value
 import za.co.absa.hyperdrive.trigger.models.UserInfo
 import org.springframework.security.core.context.SecurityContextHolder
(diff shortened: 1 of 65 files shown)


src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/UserInfoController.scala
```diff
@@ -15,19 +15,28 @@
...
+  @Value("${version:Unknown}")
+  val version: String = ""
 
   @GetMapping(path = Array("/user/info"))
   def userInfo(): UserInfo = {
     val auth = SecurityContextHolder.getContext.getAuthentication
     val principal = auth.getPrincipal.asInstanceOf[UserDetails]
-    UserInfo(principal.getUsername)
+    UserInfo(
+      username = principal.getUsername,
```
kevinwallimann: I think environment and version don't belong to userinfo. It has nothing to do with the user. Also, in principle, env and version should be displayed even if the user is logged out

jozefbakus: I am going to rewrite it

kevinwallimann (approved):
Nice!

PR #304, opened 2020-09-11
Feature/aws container poc

### LINKED ISSUE #290: Senele - Docker - Placeholder
Senele - Docker - Placeholder

senelesithole:
closes #290 

prepares hyperdrive trigger to be deployed on AWS Fargate as a docker container POC-1.

To be done POC-2 and beyond.

1. Deployment Of Hyperdrive Specific Additional jars/files.
2. Configure URSA Minor Cloud Topic Reads ( data and notification topics already onboarded).
3. Configure EMR for Spark Jobs
4. Configure And Test S3 Writes.
4. Run Full Pipeline, Kafka->Trigger( notifications or scheduled)->S3.
6. Hyperdrive Running on AWS Overall Architecture.


src/main/resources/db_scripts/db_script_latest.sql
```diff
@@ -97,12 +97,15 @@ create table "job_template" (
   "key_value_pairs" VARCHAR NOT NULL,
   "id" BIGSERIAL NOT NULL PRIMARY KEY,
   "form_config" VARCHAR NOT NULL DEFAULT 'unknown'
+<<<<<<< HEAD:src/main/resources/db_scripts/db_script_v8.sql
+=======
```
kevinwallimann: Please resolve your diffs 🤣

etc/liquibase/liquibase-maven-plugin.properties
```diff
@@ -0,0 +1,23 @@
+#
```
kevinwallimann: This file should be ignored by the `.gitignore` in the latest develop

src/main/resources/application.properties
```diff
@@ -97,4 +97,4 @@ db.connectionPool=HikariCP
 db.numThreads=4
 
 db.skip.liquibase=false
-spring.liquibase.change-log=classpath:/db_scripts/liquibase/db.changelog.yml
+spring.liquibase.change-log=classpath:/db_scripts/liquibase/db.changelog.yml
```
kevinwallimann: Please revert this change, since nothing changed except whitespace

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/WebSecurityConfig.scala
```diff
@@ -89,7 +89,8 @@ class WebSecurityConfig {
           "/scripts*.js",
           "/styles*.css",
           "/favicon.ico",
-          "/app/info"
+          "/app/info",
+          "/admin/health"
```
kevinwallimann: Why is this needed?

Dockerfile
```diff
@@ -12,13 +12,74 @@
...
+# REMOVE DEFAULT CONFIG FILES
+RUN rm -rf /etc/krb5.conf && \
+    rm -rf ${HADOOP_CONF_DIR}/* && \
+    rm -rf ${SPARK_CONF_DIR}/*
+
+# COPY REGION CONFIG FILES
+COPY install/krb/krb5.conf ${KRB_HOME}
+COPY install/hadoop/conf-ire/ ${HADOOP_CONF_DIR}
+COPY install/spark/conf-ire/ ${SPARK_CONF_DIR}
+
```
kevinwallimann: If I understand this correctly, you assume the files in the `install` folder to exist locally. And obviously you can't upload these files into this public repo. In that case, I think it doesn't make a lot of sense to keep this dockerfile here, better to move it to a private repo. The only dependency of this Dockerfile on this repo is the war file. But that's only an argument anyway, so the Dockerfile can live in another repo as well

Zejnilovic: I think you do not need a private repo. Just pass these as ENV variables and make them a path to a mounted volume and have it documented that this is expected. It then does not have to be here or in the image and is passed at runtime not at build time making the image generally usable. On our AWS you can have long-living EFS with the files.

If you hardcode these paths like this, one image can be used per environment and you would have to build a new one each time.

EDIT: I now see that what you are doing is installing Hadoop and Spark binaries and then moving config there. The same thing I talked about can be achieved by having a script as an entrypoint into the image wich would copy those configs from the mounted volume to the correct path. For Enceladus this was solved by @AdrianOlosutean, if my explanations don't make sense he could help.

kevinwallimann: I think, ultimately it depends on DevOps what works best for them. But to me, it makes no sense to have a Dockerfile that doesn't work out of the box. I would prefer a private repo that is completely self-contained.

Zejnilovic: My point of view, and I am quite vocal about this, is that an image that you upload needs to be general and "lean". It is also a best practice of Docker to not store application data/specific data in the image but on the volume. A lot of Docker images expect you to pass env variables and/or volumes to them on run time.

Also, your Dockerfile won't work out of the box even if this is a private repo. It will work for one environment only and if everything is not covered by proper DNS names, then the configs will lose their value after each restart. 

Happy to talk more offline if needed.

Dockerfile
```diff
@@ -12,13 +12,74 @@
...
+COPY --chown=0:0 install/hadoop-3.2.1 ${HADOOP_HOME}
+COPY --chown=0:0 install/spark-3.0.1-bin-hadoop3.2 ${SPARK_HOME}
+
+# REMOVE DEFAULT CONFIG FILES
+RUN rm -rf /etc/krb5.conf && \
+    rm -rf ${HADOOP_CONF_DIR}/* && \
+    rm -rf ${SPARK_CONF_DIR}/*
+
+# COPY REGION CONFIG FILES
+COPY install/krb/krb5.conf ${KRB_HOME}
```
dvagapov: I think files like krb5.conf better mount as a volume. Like Enceladus did

buildspec.yml
```diff
@@ -0,0 +1 @@
...
      - echo Building the Docker image...
      - docker build --no-cache --build-arg WAR_FILE=$WAR_FILE -t $IMAGE_REPO_NAME:$IMAGE_TAG .
      - echo Image tagging ....
      - docker tag $IMAGE_REPO_NAME:$IMAGE_TAG $AWS_ACCOUNT_ID.dkr.ecr.$AWS_DEFAULT_REGION.amazonaws.com/$IMAGE_REPO_NAME:$IMAGE_TAG
  post_build:
    commands:
      - echo Build completed on `date`
      - echo Pushing the Docker image to repository...
      - docker push $AWS_ACCOUNT_ID.dkr.ecr.$AWS_DEFAULT_REGION.amazonaws.com/$IMAGE_REPO_NAME:$IMAGE_TAG
      - echo PROCESS COMPLETE...
```
kevinwallimann: I see a lot of `^M` characters. I guess that's not as it should be?

src/main/resources/db_scripts/db_script_latest.sql
kevinwallimann: I see you resolved this, but the diff is still here. Did you forget to commit some changes?

etc/liquibase/liquibase-maven-plugin.properties
kevinwallimann: I see you resolved this, but this file is still here. Did you forget to commit some changes?

buildspec.yml
senelesithole: stranged cause I can’t pip it up from my side

senelesithole: fixed, the issue was file endings LF vs CRLF

senelesithole:
Everything should be fixed now.

kevinwallimann (approved):

PR #323, opened 2020-09-17
Feature/302 template export import

### LINKED ISSUE #302: Improve export/import for workflows with templates
In order to facilitate import / export across instances, the job templates that are referenced in a workflow, should be exported with the workflow.
See `WorkflowController.exportWorkflow`

The import-logic could be as follows:
See `WorkflowController.importWorkflow`
1. Check if job templates exist in the new environment. Match by `JobTemplate.name` (it's a unique value)
If it doesn't, fail the import

2. Replace jobTemplateId of imported workflow with the matching jobTemplateIds of the new environment and return adjusted `WorkflowJoined`

kevinwallimann:
Closes #302


jozefbakus (approved):
Nice 👍🏻

PR #328, opened 2020-09-21
Add github workflows build

kevinwallimann:
Alternative build using github actions. We can use this, while the jenkins is down or something doesn't work, like now. Maybe it's also faster than jenkins. Free quota is 500 MB storage and 2000 mins / month, so we might have to be careful not to run into the time limit  in the future.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,18 @@
+name: build
+
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


jozefbakus (approved):
Ok, lets keep jenkins and github actions. I would also leave jenkins as required condition for merge while git actions are not tested

PR #331, opened 2020-09-24
Use AsyncFlatSpec

kevinwallimann:
Sometimes, `EventProcessorTest` fails, e.g. here: https://github.com/AbsaOSS/hyperdrive-trigger/pull/330/checks?check_run_id=1159661683

I believe it might be due to not using `AsyncFlatSpec`, as it defines an execution context. But not completely sure.


jozefbakus (approved):

kevinwallimann:
Apparently, this doesn't fix the test. https://github.com/AbsaOSS/hyperdrive-trigger/pull/331/checks?check_run_id=1165542156

PR #339, opened 2020-10-05
Feature/334 bulk export

### LINKED ISSUE #334: Bulk export
The json files for the individual workflows should be packaged into a single zip file. Delivering a zip file as opposed to one large json file with an array of workflows allows the user to easier exclude workflows from the zip file in the import. The user might want to exclude workflows to try a few workflows before importing a big batch of workflows or exclude workflows that cause import errors (workflows that violate validations)

Action should complete fully or not at all.

kevinwallimann:
Closes #334

### DIFF
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/WorkflowController.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/WorkflowController.scala
@@ -28,7 +28,7 @@ import org.springframework.web.multipart.MultipartFile
 import za.co.absa.hyperdrive.trigger.ObjectMapperSingleton
 import za.co.absa.hyperdrive.trigger.api.rest.services.WorkflowService
 import za.co.absa.hyperdrive.trigger.models._
-import za.co.absa.hyperdrive.trigger.models.errors.{ApiError, ApiException}
+import za.co.absa.hyperdrive.trigger.models.errors.{ApiError, ApiException, GenericError}
 
 import scala.compat.java8.FutureConverters._
 import scala.concurrent.ExecutionContext.Implicits.global
@@ -114,7 +114,11 @@ class WorkflowController @Inject()(workflowService: WorkflowService) {
   @GetMapping(path = Array("/workflows/export"))
   def exportWorkflows(@RequestParam jobIds: Array[Long]): CompletableFuture[ResponseEntity[ByteArrayResource]] = {
     workflowService.exportWorkflows(jobIds).map { exportItems =>
-      if (exportItems.size == 1) {
+      if (exportItems.isEmpty) {
+        val jobIdsString = jobIds.map(_.toString).reduce(_ + ", " + _)
+        throw new ApiException(GenericError(s"The requested workflows with ids $jobIdsString don't exist."))
+      }
+      else if (exportItems.size == 1) {
         exportWorkflowAsJson(exportItems.head)
       } else {
         exportWorkflowsAsZip(exportItems)
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/WorkflowController.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/WorkflowController.scala
@@ -17,16 +17,16 @@ package za.co.absa.hyperdrive.trigger.api.rest.controllers
 
 import java.util.concurrent.CompletableFuture
 
-import com.fasterxml.jackson.databind.annotation.JsonDeserialize
 import javax.inject.Inject
+import org.springframework.beans.factory.annotation.Value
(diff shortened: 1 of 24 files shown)


src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/WorkflowController.scala
```diff
@@ -106,18 +109,28 @@ class WorkflowController @Inject()(workflowService: WorkflowService) {
...
-        ObjectMapperSingleton.getObjectMapper.writerWithDefaultPrettyPrinter.writeValueAsBytes(workflowExport)
-      )
-
-      ResponseEntity.ok()
-        .contentType(MediaType.parseMediaType("application/json"))
-        .header(HttpHeaders.CONTENT_DISPOSITION, s"attachment; filename=${workflowExport.workflowJoined.name}.json")
-        .body(resource)
-    }.toJava.toCompletableFuture
+  @GetMapping(path = Array("/workflows/export"))
+  def exportWorkflows(@RequestParam jobIds: Array[Int]): CompletableFuture[ResponseEntity[ByteArrayResource]] = {
```
jozefbakus: jobIds should be array of Long

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/WorkflowController.scala
```diff
@@ -106,18 +109,28 @@ class WorkflowController @Inject()(workflowService: WorkflowService) {
...
+    val jobIdsLong = jobIds.map(_.toLong).toSeq
+    if (jobIdsLong.size > 1) {
+      workflowService.exportWorkflowsAsZip(jobIdsLong).map(baos =>
+        ResponseEntity.ok()
+          .contentType(MediaType.parseMediaType("application/zip"))
+          .header(HttpHeaders.CONTENT_DISPOSITION, s"attachment; filename=workflows${environment}.zip")
```
jozefbakus: What about dash ? workflows-${environment}.zip

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/WorkflowController.scala
jozefbakus: I would create two endpoints, one for single workflow and one for multiple workflows.
Here is one function which does two things. 
But leave it as it is.

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/WorkflowService.scala
```diff
@@ -191,6 +196,26 @@ class WorkflowServiceImpl(override val workflowRepository: WorkflowRepository,
     } yield WorkflowImportExportWrapper(workflow, jobTemplates)
   }
 
+  override def exportWorkflowsAsZip(workflowIds: Seq[Long])(implicit ec: ExecutionContext): Future[ByteArrayOutputStream] = {
+    val workflowExportsFut = Future.sequence(workflowIds.map(exportWorkflow(_)))
```
jozefbakus: I would rather do it with one call to db. (or two, one for workflows and one for templates)

jozefbakus (changes requested):

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/WorkflowController.scala
kevinwallimann: ok

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/WorkflowController.scala
kevinwallimann: I could extract the parts into two functions. I think two endpoints aren't necessary. If you would call `/workflows/export/42` you would get a json wrapped in a zip file, while `workflow/export/42` would return only the json. I think it doesn't make sense to have two ways of exporting.

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/WorkflowService.scala
kevinwallimann: ok

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/WorkflowController.scala
kevinwallimann: Ok.

jozefbakus: OK

jozefbakus (approved):
👍🏻

PR #341, opened 2020-10-08
Feature/340 remove future either

### LINKED ISSUE #340: Remove usage of Future[Either[... in WorkflowService
Currently, the following methods return an `Either[ApiError, T]` wrapped by a `Future`:
- `WorkflowService.createWorkflow`
- `WorkflowService.updateWorkflow`
- `WorkflowRepository.insertWorkflow`
- `WorkflowRepository.updateWorkflow`

For these methods, this is overcomplicated and could be replaced by returning a simple `Future[T]` and returning a `Future.failed(new ApiException(apiErrors))` in case of errors. 
The difference between returning a failed future or an `Either` wrapped by a `Future` is that the latter always returns a successful future, i.e. does not fail fast and subsequent futures can be executed.
The current implementation would work with fail-fast semantics on `WorkflowRepository.insertWorkflow` and `WorkflowRepository.updateWorkflow` because they are the last methods of `WorkflowService.createWorkflow` and `WorkflowService.updateWorkflow` anyway

kevinwallimann:
Closes #340

### DIFF
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/WorkflowService.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/WorkflowService.scala
@@ -18,8 +18,8 @@ package za.co.absa.hyperdrive.trigger.api.rest.services
 import org.springframework.security.core.context.SecurityContextHolder
 import org.springframework.security.core.userdetails.UserDetails
 import org.springframework.stereotype.Service
-import za.co.absa.hyperdrive.trigger.models.errors.{ApiError, GenericError}
 import za.co.absa.hyperdrive.trigger.models._
+import za.co.absa.hyperdrive.trigger.models.errors.{ApiException, GenericError}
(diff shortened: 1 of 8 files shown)


src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/WorkflowService.scala
```diff
@@ -87,46 +79,45 @@ class WorkflowServiceImpl(override val workflowRepository: WorkflowRepository,
...
+      ),
+      dagDefinitionJoined = workflow.dagDefinitionJoined.copy(
+        id = originalWorkflow.dagDefinitionJoined.id,
+        workflowId = originalWorkflow.id
+      )
+    )
+  }
+
   override def switchWorkflowActiveState(id: Long)(implicit ec: ExecutionContext): Future[Boolean] = {
-    val userName = getUserName.apply();
```
jozefbakus, line 129: Thanks for this :D

jozefbakus (approved):
👍🏻

PR #370, opened 2021-01-21
Feature/368 scalable scheduler

### LINKED ISSUE #368: Make scheduler scalable
Currently, the scheduler can only run on a single node. Running the scheduler on multiple instances would result in duplicate job submissions for all workflows.

To enable scalability (and HA), workflows should be distributed among available instances.

kevinwallimann:
Closes #368 
# Solution description
- The load will be distributed in a coarse manner, by distributing `Workflow`s among instances. Each instance will then only poll sensors and execute dags for the workflows that are assigned to it. This is not a perfect distribution, since some workflows may have multiple dags, and not all sensor types are equally expensive.
- Application instances can be started and stopped at any time, without knowing each other. The DB acts as a coordinator and keeps track of active instances.

## Coordination of instances
The new table `scheduler_instance` keeps track of active instances and contains three columns: id, status (Active, Deactivated), and heartbeat. Each instance inserts an entry when it starts, and subsequently updates the heartbeat with the current timestamp in every iteration of the scheduler. If the heartbeat of an instance lags behind a certain time (`scheduler.lag.threshold`), its status will be updated to `Deactivated` by any of the other instances. If an instance was deactivated prematurely (e.g. due to network delay etc.), it will stop (`JobScheduler.stopManager`) itself. If an instance was deactivated, it will be deleted from the table after deleting its references on the workflow table. Thus, the `scheduler_instance` table does not grow indefinitely and should stay very small.

## Distribution of workflows
See `WorkflowBalancer` and `WorkflowBalancingService`
- The column `scheduler_instance_id` was added to the table `workflow`, which is a foreign key to `scheduler_instance` and is nullable. An important invariant is that one workflow may never be assigned to more than one instance at any time. However, it is possible that a workflow may be unassigned for a certain time.
- The workflow assignment is computed by each instance locally according to this: workflow_id mod #active_instances == instance_rank. instance_rank is the rank of the instance id
- Workflows of running dags will be retained on the instance and are only subject to rebalancing when they have completed.
- `scheduler_instance_id` of a workflow is always set to `null` by the instance that releases it, and workflows can only be acquired if the `scheduler_instance_id` is `null`. This ensures no workflows will be held by two instances at the same time.
- A workflow may be temporarily unassigned during rebalancing phases, namely when either the number of instances or the number of workflows is observed differently on each instance. All workflows should eventually be assigned when the observed instances and workflows are the same across all instances

### Efficiency
- At every iteration, each instance will perform an update and a select query to `scheduler_instance`, and also select the maximal id of `workflow` (The latter could be optimized by querying the underlying sequence instead if needed). See `WorkflowBalancer`
- If a rebalance is required, more expensive queries are executed, most notably selecting all workflows, as well as update queries on workflows. Rebalances only need to take place, if an instance is added or deactivated, or if a workflow is added (No rebalance is carried out when a workflow is removed), or if the desired workflows could not yet be acquired, because another instance is still holding it.

### Pitfalls
- Other instances may still hold a workflow because the dag is still running. In case a rebalance is triggered due to changes in instances or workflows, while a dag is submitted (can take several minutes), it will try to rebalance during every iteration until the dag is submitted and "released". This could cause higher SQL execution times due to the relatively expensive queries in the rebalance phase.
- If a rebalance happens exactly when a time-based workflow should be executed, the execution might be missed. E.g. instance 1 drops the workflow at xx:27, the workflow is due at xx:30 and instance 2 only acquires the workflow at xx:32. This could be fixed by using the clustered version of Quartz.


src/main/resources/db_scripts/delta_script_v9_to_v10.sql
```diff
@@ -0,0 +1,30 @@
+/*
```
jozefbakus: What about db_script_v10.sql ?

src/main/scala/za/co/absa/hyperdrive/trigger/models/tables/WorkflowTable.scala
```diff
@@ -30,17 +30,19 @@ trait WorkflowTable {
     def project: Rep[String] = column[String]("project")
     def created: Rep[LocalDateTime] = column[LocalDateTime]("created")
     def updated: Rep[Option[LocalDateTime]] = column[Option[LocalDateTime]]("updated")
+    def schedulerInstanceId: Rep[Option[Long]] = column[Option[Long]]("scheduler_instance_id")
     def id: Rep[Long] = column[Long]("id", O.PrimaryKey, O.AutoInc, O.SqlType("BIGSERIAL"))
 
-    def * : ProvenShape[Workflow] = (name, isActive, project, created, updated, id) <> (
+    def * : ProvenShape[Workflow] = (name, isActive, project, created, updated, schedulerInstanceId, id) <> (
```
jozefbakus, line 39: Can you add ForeignKeyQuery here for shedulerInstanceId ?

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/JobScheduler.scala
```diff
@@ -44,19 +49,29 @@ class JobScheduler @Inject()(sensors: Sensors, executors: Executors, dagInstance
   private var runningScheduler: Future[Unit] = Future.successful((): Unit)
   private var runningSensors = Future.successful((): Unit)
   private var runningEnqueue = Future.successful((): Unit)
-  private val runningDags = mutable.Map.empty[Long, Future[Unit]]
+  private val runningDags = mutable.Map.empty[RunningDagsKey, Future[Unit]]
 
   def startManager(): Unit = {
```
jozefbakus, line 55: Logic of this method is fine, but  it can create multiple futures (assignedWorkflowsFut) and code inside accesses variables and resources

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/cluster/WorkflowBalancingService.scala
```diff
@@ -0,0 +1,84 @@
...
+    val myRank = getRank(activeInstances, myInstanceId)
+    logger.info(s"Rebalancing workflows on scheduler instance id = $myInstanceId, rank = $myRank," +
+      s" active instance ids = ${activeInstances.map(_.id).sorted}, retaining workflow ids = ${runningWorkflowIds}")
+    for {
+      (releasedWorkflowsCount, instancesDeletedCount) <- workflowRepository.releaseWorkflowAssignmentsOfDeactivatedInstances()
+      _ = if (releasedWorkflowsCount > 0) {
+        logger.info(s"Scheduler instance id = $myInstanceId released $releasedWorkflowsCount workflows of " +
+          s"$instancesDeletedCount deactivated instances")
+      }
+      allWorkflows <- workflowRepository.getWorkflows()
```
jozefbakus, line 51: Quite an expensive query.

src/test/scala/za/co/absa/hyperdrive/trigger/persistance/SchedulerInstanceRepositoryTest.scala
```diff
@@ -0,0 +1,94 @@
...
+    allInstances should contain allElementsOf TestData.schedulerInstances
+  }
+
+  "updateHeartbeat" should "update the last heartbeat of an active instance" in {
+    val now = LocalDateTime.now()
+    val result = await(schedulerInstanceRepository.updateHeartbeat(11))
+    val updatedInstance = await(db.run(schedulerInstanceTable.filter(_.id === 11L).result.head))
+
+    result shouldBe 1
+    updatedInstance.lastHeartbeat.isBefore(now) shouldBe false
```
jozefbakus: Should not be  .... positively formulated test is easier to read 
    updatedInstance.lastHeartbeat.isAfter(now) shouldBe true

jozefbakus:
1. When workflow is created it gets scheduler_instance_id, but when you update workflow, scheduler_instance_id is deleted and stays null
2. Is possible to have different distribution logic? 
- Let's say we have only even workflow ids, in that case all workflows will be only on one instance
- Not distributing inactive workflows ?

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/cluster/WorkflowBalancingService.scala
kevinwallimann: Yes in the sense that it fetches all workflows. But it's not the joined workflow, i.e. it only selects name, project, isActive, etc. I'd argue it's probably not a performance bottleneck, but I can easily optimize the query by only selecting the ids (and omitting the sort)

src/test/scala/za/co/absa/hyperdrive/trigger/persistance/SchedulerInstanceRepositoryTest.scala
kevinwallimann: Haha, yes I know. What I really wanted to use was `isAfterOrEqual`, but that doesn't exist unfortunately. But I'll do something about it

kevinwallimann:
> 1. When workflow is created it gets scheduler_instance_id, but when you update workflow, scheduler_instance_id is deleted and stays null
> 2. Is possible to have different distribution logic?
> 
> * Let's say we have only even workflow ids, in that case all workflows will be only on one instance
> * Not distributing inactive workflows ?

1. Hmm, that's not the right behavior. At creation, it should be null, and at update, it shouldn't change the scheduler instance id. I'll fix that
2. 
- It's quite unrealistic to only have even workflow ids. You'd have to delete every second workflow. In such a case, I could locally map workflowIds to ranks (min workflow Id -> 0, next higher id -> 1, and so on) and distribute according to ranks. It would then not be enough to invoke the redistribution only if a new workflow is detected, but it would also need to be invoked whenever a workflow is deleted, which probably requires to get all workflow ids in every scheduler iteration.
- This could be done, but then the redistribution needs to be invoked whenever the active state of any workflow changes. Therefore I have to query the active state of all workflows at every scheduler iteration.

Both optimizations are a trade-off. For the initial implementation I chose an approach that is as simple as possible. It's quite likely that there will be some skew in the actual load (because of deleted workflows, and potential imbalances in active workflows), but we'll have to see how bad the imbalances will be.

src/main/resources/db_scripts/delta_script_v9_to_v10.sql
kevinwallimann: Done. `db_script_latest`.sql

src/main/scala/za/co/absa/hyperdrive/trigger/models/tables/WorkflowTable.scala
kevinwallimann: Done

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/JobScheduler.scala
kevinwallimann: Done

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/cluster/WorkflowBalancingService.scala
kevinwallimann: Do you think it's worthwhile to only select the two instead of the 6 columns of workflow? Otherwise, I won't do it, because it's not worth the complication.

src/test/scala/za/co/absa/hyperdrive/trigger/persistance/SchedulerInstanceRepositoryTest.scala
kevinwallimann: Done.

src/main/resources/db_scripts/liquibase/v0.3.3.add-scheduler-instances.yml
```diff
@@ -0,0 +1,25 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+databaseChangeLog:
+  - changeSet:
+      id: v0.3.3.add-scheduler-instances
```
jozefbakus: This change will be in the v0.3.4 release

src/main/scala/za/co/absa/hyperdrive/trigger/persistance/WorkflowRepository.scala
```diff
@@ -193,9 +193,12 @@ class WorkflowRepositoryImpl(override val workflowHistoryRepository: WorkflowHis
   }
 
   override def updateWorkflow(workflow: WorkflowJoined, user: String)(implicit ec: ExecutionContext): Future[Unit] = {
+    val w = workflow.toWorkflow.copy(updated = Option(LocalDateTime.now()))
     db.run(
       (for {
-        w <- workflowTable.filter(_.id === workflow.id).update(workflow.toWorkflow.copy(updated = Option(LocalDateTime.now())))
+        w <- workflowTable.filter(_.id === workflow.id)
+          .map(t => (t.name, t.isActive, t.project, t.updated))
```
jozefbakus, line 200: I don't like this approach but I couldn't find a better way.

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/cluster/WorkflowBalancingService.scala
jozefbakus: I am not sure if it is really worth but I would do it. 
+  acquiredWorkflows <- workflowRepository.getWorkflowsBySchedulerInstance(myInstanceId) - is it really needed ? I think you can return workflowIdsToAcquire

jozefbakus:
> > 1. When workflow is created it gets scheduler_instance_id, but when you update workflow, scheduler_instance_id is deleted and stays null
> > 2. Is possible to have different distribution logic?
> > 
> > 
> > * Let's say we have only even workflow ids, in that case all workflows will be only on one instance
> > * Not distributing inactive workflows ?
> 
> 1. Hmm, that's not the right behavior. At creation, it should be null, and at update, it shouldn't change the scheduler instance id. I'll fix that
> 
> * It's quite unrealistic to only have even workflow ids. You'd have to delete every second workflow. In such a case, I could locally map workflowIds to ranks (min workflow Id -> 0, next higher id -> 1, and so on) and distribute according to ranks. It would then not be enough to invoke the redistribution only if a new workflow is detected, but it would also need to be invoked whenever a workflow is deleted, which probably requires to get all workflow ids in every scheduler iteration.
> * This could be done, but then the redistribution needs to be invoked whenever the active state of any workflow changes. Therefore I have to query the active state of all workflows at every scheduler iteration.
> 
> Both optimizations are a trade-off. For the initial implementation I chose an approach that is as simple as possible. It's quite likely that there will be some skew in the actual load (because of deleted workflows, and potential imbalances in active workflows), but we'll have to see how bad the imbalances will be.

1. Works fine now
2. I do agree 

**Closed**

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/cluster/WorkflowBalancingService.scala
kevinwallimann: No, because `acquireWorkflowAssignments` does 
```
UPDATE workflow SET scheduler_instance_id = myInstanceId 
WHERE id IN (workflowIdsToAcquire) 
AND scheduler_instance_id IS NULL
```

So not all `workflowIdsToAcquire` might be updated, therefore 
```
SELECT id FROM workflow 
WHERE scheduler_instance_id = myInstanceId
```
 is not necessarily the same as `workflowIdsToAcquire`

src/main/scala/za/co/absa/hyperdrive/trigger/persistance/WorkflowRepository.scala
kevinwallimann: I have the same sentiment. I was thinking of creating a `WorkflowUpdate` case class and somehow register this to the table, but you'd still need to change multiple places if you want to add a new field in `Workflow`. Let's just never add a field to `Workflow` anymore :stuck_out_tongue:

jozefbakus (approved):
Very nice

PR #371, opened 2021-01-22
Feature/liquibase 6

kevinwallimann:
# How to use
## Development

1. Copy `/etc/liquibase/liquibase-maven-plugin.properties.template` to `/etc/liquibase/liquibase-maven-plugin.properties` and modify it as needed (db url, user, password).

2. In the root folder of the project, execute `mvn liquibase:status`.
ATTENTION :bomb: :boom: :skull_and_crossbones: `mvn liquibase:dropAll` will erase your database. You should export all test workflows before, to import them after dropAll

`mvn liquibase:status` View status
`mvn liquibase:dropAll` Drops all tables, views etc.
`mvn liquibase:update` Applies all pending changesets

### Adding new changesets
1. In the `src/main/resources/db_scripts/liquibase` folder, for every new changeset, add a SQL and yml file (copy paste from existing) Unfortunately, some redundancy w.r.t. filenames is required in the yml file, so pay attention to it.
2. Add the changeset in `db.changelog.yml`. If it's not listed there, it won't be applied. Don't forget this step.

### Initial liquibase tables generation on existing database
See `liquibase-init.sql`
- How this was generated: To initially generate the liquibase tables `databasechangelog` and `databasechangeloglock`, do the following:
Execute `mvn liquibase:changelogSyncSQL`. This generates the SQL needed to create the changelog table, where liquibase tracks the db changesets and adds rows for changesets that have already been applied and are now tracked in hindsight. Any inserts that have not been applied yet have to be removed before applying the sql to the db.

#### Automatically on startup
On startup, `LiquibaseConfiguration` automatically checks if all pending changesets have been applied. If not, the schema is updated with the pending changesets. 
If there is a SQL error or an error with liquibase itself, an error will be thrown and the app will not start. If such errors should be ignored and the application started anyway, set `db.skip.liquibase` to `true` in the application properties before starting the app.

#### Manually using maven liquibase plugin
You should probably not do this. Make sure your db url, user and password are correct! Make sure you checked out the correct branch!
`mvn liquibase:status` to see which changesets will be applied
`mvn liquibase:update` to apply the changesets

Alternatively, you could manually apply the changes, and then manually insert a corresponding entry in the `databasechangelog` table. Again, not recommended.


README.md
```diff
@@ -202,6 +204,13 @@ Hyperdrive-trigger can be packaged as a Web Application Archive and executed in
 - Without tests: `mvn clean package -DskipTests`
 - With unit tests: `mvn clean package`
 
+## Liquibase
+The liquibase maven plugin may be used to issue liquibase commands. To use it, copy 
+`/etc/liquibase/liquibase-maven-plugin.properties.template` to `/etc/liquibase/liquibase-maven-plugin.properties` and modify it as needed.
+
+Then, the liquibase maven plugin can be executed, e.g.
+`mvn liquibase:status`
```
jozefbakus: I would include also mvn liquibase:dropAll and mvn liquibase:update

pom.xml
```diff
@@ -348,6 +364,22 @@
                     </archive>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.liquibase</groupId>
+                <artifactId>liquibase-maven-plugin</artifactId>
+                <version>${liquibase-maven-plugin.version}</version>
+                <configuration>
+                    <propertyFile>etc/liquibase/liquibase-maven-plugin.properties</propertyFile>
```
jozefbakus, line 372: Is it possible to include properties to application.properties with prefix ? 

`db.driver=net.bull.javamelody.JdbcDriver
db.url=jdbc:postgresql://localhost:5432/hyperdriver?driver=org.postgresql.Driver
db.keepAliveConnection=true
db.connectionPool=HikariCP
db.numThreads=4

liquibase.db.driver = org.postgresql.Driver
liquibase.db.url = ${db.url}
`

src/main/resources/db_scripts/liquibase/liquibase-init.sql
```diff
@@ -0,0 +1,60 @@
...
+
+INSERT INTO databasechangelog (ID, AUTHOR, FILENAME, DATEEXECUTED, ORDEREXECUTED, MD5SUM, DESCRIPTION, COMMENTS, EXECTYPE, CONTEXTS, LABELS, LIQUIBASE, DEPLOYMENT_ID) VALUES ('v0.2.0.add-job-template', 'kevin.wallimann@absa.africa', 'v0.2.0.add-job-template', NOW(), 6, '8:ac942942e1af37b9f3b51640941bfa0f', 'sqlFile', '', 'EXECUTED', 'default', NULL, '3.10.2', '1336601767');
+
+INSERT INTO databasechangelog (ID, AUTHOR, FILENAME, DATEEXECUTED, ORDEREXECUTED, MD5SUM, DESCRIPTION, COMMENTS, EXECTYPE, CONTEXTS, LABELS, LIQUIBASE, DEPLOYMENT_ID) VALUES ('v0.2.0.delete-job-type', 'kevin.wallimann@absa.africa', 'v0.2.0.delete-job-type', NOW(), 7, '8:dc26d32dc134c39cbc9c3e990d9fb99c', 'sqlFile', '', 'EXECUTED', 'default', NULL, '3.10.2', '1336601767');
+
+INSERT INTO databasechangelog (ID, AUTHOR, FILENAME, DATEEXECUTED, ORDEREXECUTED, MD5SUM, DESCRIPTION, COMMENTS, EXECTYPE, CONTEXTS, LABELS, LIQUIBASE, DEPLOYMENT_ID) VALUES ('v0.2.0.add-triggered-by', 'kevin.wallimann@absa.africa', 'v0.2.0.add-triggered-by', NOW(), 8, '8:20947f500206c9f3dcbced94d9079582', 'sqlFile', '', 'EXECUTED', 'default', NULL, '3.10.2', '1336601767');
```
jozefbakus: AUTHOR should be HyperdriveDevTeam@absa.africa ? 
I don't mind leaving it this way

pom.xml
kevinwallimann: Hmm, I don't think the liquibase maven plugin can deal with prefixes, unfortunately. Also, I didn't want to put the liquibase prefixes into the `application.properties` because the `LiquibaseConfiguration` directly uses `db.url` etc. so it might be misleading. E.g. you could configure `liquibase.db.url=B` and `db.url=A`, then `LiquibaseConfiguration` would just take A and ignore B.

I don't find the duplication of the db configuration that bad, because there should be no need to use the liquibase maven plugin for anything else than your local development DB

src/main/resources/db_scripts/liquibase/liquibase-init.sql
kevinwallimann: Hmm. The important thing is to track who made the change. To that end, the personal email is better. However, you can find out the author using the filename and git blame as well. So we can also use the group email address so as not to write down our personal emails into the DB

kevinwallimann: I changed it

README.md
kevinwallimann: Done

pom.xml
jozefbakus: I am fine with that. 
I didn't notice that LiquibaseConfiguration loads configuration from application.properties.

jozefbakus (dismissed):

jozefbakus (approved):

senelesithole (approved):
This is an awesome plugin, I love it. One thing, I wish the hint " **You may want to import / export all test workflows, to recreate them after dropAl**l" was at the top of the paragraph , I have lost all my test workflow data now 😔. 

I like how it shows/tells you that you have an unapplied changeset, work hand in hand with git. 👌🏾

PR #386, opened 2021-03-15
Feature/380 merge extra java options

### LINKED ISSUE #380: Merge extraJavaOptions from template, job and properties file
**Current situation**
With #366 extraJavaOptions from the job definition and options globally defined in the properties file are merged. However, it is also possible to define extraJavaOptions in job templates. Currently, extraJavaOptions defined in a job template are dropped if extraJavaOptions are defined in the job definition.
image

image

image

**Desired behaviour**
If a job template defines additional spark config keys `spark.executor.extraJavaOptions` or `spark.driver.extraJavaOptions`, the value from either property should not be overwritten by the job definition in any case, but merged with the `spark.executor.extraJavaOptions` and `spark.driver.extraJavaOptions` of the job definition.

The values should be concatenated. The job definition's extraJavaOptions should be appended to the job template's extraJavaOptions such that anything defined in the job definition's extraJavaOptions has precedence over the job template.

kevinwallimann:
Closes #380

### DIFF
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/spark/SparkExecutor.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/spark/SparkExecutor.scala
@@ -30,13 +30,13 @@ import play.api.libs.ws.ahc.StandaloneAhcWSClient
 import za.co.absa.hyperdrive.trigger.models.enums.JobStatuses._
 import za.co.absa.hyperdrive.trigger.scheduler.executors.Executor
 import za.co.absa.hyperdrive.trigger.scheduler.utilities.SparkExecutorConfig
+import za.co.absa.hyperdrive.trigger.scheduler.utilities.JobDefinitionConfig.{KeysToMerge, MergedValuesSeparator}
 import play.api.libs.ws.JsonBodyReadables._
 import za.co.absa.hyperdrive.trigger.scheduler.executors.spark.{FinalStatuses => YarnFinalStatuses}
 import org.slf4j.LoggerFactory
 
 object SparkExecutor extends Executor {
   private val wsClient = StandaloneAhcWSClient()(ActorMaterializer()(ActorSystem()))
-  private val keysToMerge = Set("spark.executor.extraJavaOptions", "spark.driver.extraJavaOptions")
 
   override def execute(jobInstance: JobInstance, updateJob: JobInstance => Future[Unit])
                       (implicit executionContext: ExecutionContext): Future[Unit] = {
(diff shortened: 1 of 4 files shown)


src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/utilities/Configs.scala
```diff
@@ -100,4 +100,9 @@ object SparkExecutorConfig {
     Try(Configs.conf.getString("sparkYarnSink.filesToDeploy").split(",").toSeq).getOrElse(Seq.empty[String])
   def getAdditionalConfs: Map[String, String] =
     Configs.getMapFromConf("sparkYarnSink.additionalConfs")
+}
+
+object JobDefinitionConfig {
+  val KeysToMerge = Set("spark.executor.extraJavaOptions", "spark.driver.extraJavaOptions")
```
jozefbakus, line 106: I would move this to SparkExecutorConfig

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/spark/SparkExecutor.scala
```diff
@@ -92,17 +92,17 @@ object SparkExecutor extends Executor {
     sparkParameters.additionalJars.foreach(additionalJar => sparkLauncher.addJar(additionalJar))
     sparkParameters.additionalFiles.foreach(additionalFile => sparkLauncher.addFile(additionalFile))
     sparkParameters.additionalSparkConfig.foreach(conf => sparkLauncher.setConf(conf._1, conf._2))
-    mergeAdditionalSparkConfig(SparkExecutorConfig.getAdditionalConfs, sparkParameters.additionalSparkConfig, keysToMerge)
+    mergeAdditionalSparkConfig(SparkExecutorConfig.getAdditionalConfs, sparkParameters.additionalSparkConfig)
```
jozefbakus, line 95: Does this override configs from lines above ? 
    SparkExecutorConfig.getAdditionalConfs.foreach(conf => sparkLauncher.setConf(conf._1, conf._2))
and
    sparkParameters.additionalSparkConfig.foreach(conf => sparkLauncher.setConf(conf._1, conf._2))

src/test/scala/za/co/absa/hyperdrive/trigger/api/rest/utils/JobTemplateResolutionUtilTest.scala
```diff
@@ -170,6 +170,35 @@ class JobTemplateResolutionUtilTest extends FlatSpec with Matchers {
     )
```
jozefbakus, line 170: Hmm is this behavior correct ? It is the same for (JobTemplate and SparkExecutor)
additionalSparkConfig: "spark.driver.extraJavaOptions" -> "-user.prop=userDriver -aaa.bbb=xxx"
templateParameters: "spark.driver.extraJavaOptions" -> "-template.prop=templateDriver -aaa.bbb=yyy"
Result is:
"spark.driver.extraJavaOptions" -> "-template.prop=templateDriver -aaa.bbb=yyy -user.prop=userDriver -aaa.bbb=xxx"

kevinwallimann: Yes, it's the desired behavior. Both the template properties and the user properties should be included in the final extraJavaOptions. If you want to override a template option, the assumption is that for java options, the succeeding options override the preceding options, that's why the user options are after the template options

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/spark/SparkExecutor.scala
kevinwallimann: Yes, but it only overrides the `KeysToMerge` configs. I could also filter out the `KeysToMerge` in `sparkParameters.additionalSparkConfig.foreach`, but that would mainly just be for better readability, the end result wouldn't change.

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/utilities/Configs.scala
kevinwallimann: Yes, but it's not only about the SparkExecutor. It's also about the JobTemplate and JobDefinition. But anyway I can move it

jozefbakus: Yes you are right. Leave it here

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/spark/SparkExecutor.scala
jozefbakus: Yes, I thought so. You can leave it as it is.

jozefbakus (approved):

PR #390, opened 2021-03-16
refactor get changed sensors query

### LINKED ISSUE #388: Find changed sensors - too big query
Find changed sensors - too big query:
In class SensorRepository method getChangedSensors builds big query where all sensor parameters are compared. 
Updated field on workflow table should be used

jozefbakus:
Implements: #388 
New property in application.properties: scheduler.sensors.changedSensorsMinimalChunkQuerySize

### DIFF
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -49,6 +49,7 @@ scheduler.executors.thread.pool.size=30
 scheduler.jobs.parallel.number=100
 scheduler.heart.beat=5000
 scheduler.lag.threshold=20000
+scheduler.sensors.changedSensorsMinimalChunkQuerySize=100
 
 #Kafka sensor properties.
 kafkaSource.group.id=hyper_drive_${appUniqueId}
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/persistance/SensorRepository.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/persistance/SensorRepository.scala
@@ -54,7 +54,7 @@ class SensorRepositoryImpl extends SensorRepository {
     Future.sequence(originalSensors.grouped(100).toSeq.map(group => getChangedSensorsInternal(group))).map(_.flatten)
   }
 
-  def getChangedSensorsInternal(originalSensors: Seq[Sensor])(implicit ec: ExecutionContext): Future[Seq[Sensor]] = db.run {(
+  private def getChangedSensorsInternal(originalSensors: Seq[Sensor])(implicit ec: ExecutionContext): Future[Seq[Sensor]] = db.run {(
     for {
       sensor <- sensorTable if originalSensors
         .map(originalSensor => sensorIsDifferent(sensor, originalSensor))
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/persistance/SensorRepository.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/persistance/SensorRepository.scala
@@ -17,6 +17,7 @@ package za.co.absa.hyperdrive.trigger.persistance
 
 import org.springframework.stereotype
 import za.co.absa.hyperdrive.trigger.models.Sensor
+import za.co.absa.hyperdrive.trigger.scheduler.utilities.SensorsConfig
 
 import scala.concurrent.{ExecutionContext, Future}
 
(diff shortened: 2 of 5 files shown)


src/test/scala/za/co/absa/hyperdrive/trigger/persistance/SensorRepositoryTest.scala
```diff
@@ -122,6 +123,25 @@ class SensorRepositoryTest extends FlatSpec with Matchers with BeforeAndAfterAll
       .map { case (sensor, _) => sensor }
   }
 
+  "sensorRepository.getChangedSensors" should "not fail on large number of sensors" in {
```
kevinwallimann, line 126: Does this test actually fail without the fix?

src/main/scala/za/co/absa/hyperdrive/trigger/persistance/SensorRepository.scala
```diff
@@ -50,7 +50,11 @@ class SensorRepositoryImpl extends SensorRepository {
     }).result
   }
 
-  override def getChangedSensors(originalSensors: Seq[Sensor])(implicit ec: ExecutionContext): Future[Seq[Sensor]] = db.run {(
+  override def getChangedSensors(originalSensors: Seq[Sensor])(implicit ec: ExecutionContext): Future[Seq[Sensor]] = {
+    Future.sequence(originalSensors.grouped(100).toSeq.map(group => getChangedSensorsInternal(group))).map(_.flatten)
```
kevinwallimann: `100` is a magic constant. I'd move it to `Configs.scala`

src/test/scala/za/co/absa/hyperdrive/trigger/persistance/SensorRepositoryTest.scala
jozefbakus: Yes

src/main/scala/za/co/absa/hyperdrive/trigger/persistance/SensorRepository.scala
jozefbakus: Ok

jozefbakus: Done

src/test/scala/za/co/absa/hyperdrive/trigger/persistance/SensorRepositoryTest.scala
kevinwallimann: Nice

src/main/resources/application.properties
```diff
@@ -49,6 +49,7 @@ scheduler.executors.thread.pool.size=30
 scheduler.jobs.parallel.number=100
 scheduler.heart.beat=5000
 scheduler.lag.threshold=20000
+scheduler.sensors.changedSensorsMinimalChunkQuerySize=100
```
kevinwallimann, line 52: I'd have been ok with the value hard-coded in the configs, just not in the service itself, but anyway, now you've already done it.
It seems that we don't use camelCase in the properties files. Almost all properties are lower case with a dot between words. The only exception is javamelody which uses kebab case

kevinwallimann: Taking this back. camelCase is still better than dots, since the dots imply a hierarchy

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/utilities/Configs.scala
```diff
@@ -69,6 +69,8 @@ object KafkaConfig {
 object SensorsConfig {
   def getThreadPoolSize: Int =
     Configs.conf.getInt("scheduler.sensors.thread.pool.size")
+  def getChangedSensorsMinimalChunkQuerySize: Int =
```
kevinwallimann: As in the other PR, I think we should use `val` instead of `def` to fail fast

src/main/resources/application.properties
jozefbakus: Ok

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/utilities/Configs.scala
jozefbakus: ok

kevinwallimann (approved):
I hope this doesn't create a bunch of false-positive merge conflicts :)

kevinwallimann:
Sorry, haven't seen this before. Did you forget converting this into a `val`: `def getConsumerProperties`?

```
object KafkaConfig {
  def getConsumerProperties(kafkaSettings: KafkaSettings): Properties = {
    val properties = new Properties()
```

PR #391, opened 2021-03-16
Feature/385 symlinks for files

### LINKED ISSUE #385: Symlink for files
All files executed by trigger(job jars and other job executables) should be in one folder.
Path to the folder should be defined in the application properties file.
PathToFolder + JobExecutableName(used in workflows/jobs) should be used by job executor.

jozefbakus:
closes #385 

new property in application properties: scheduler.executors.executablesFolder

### DIFF
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/utilities/Configs.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/utilities/Configs.scala
@@ -69,8 +69,8 @@ object KafkaConfig {
 object SensorsConfig {
   val getThreadPoolSize: Int =
     Configs.conf.getInt("scheduler.sensors.thread.pool.size")
-  val getChangedSensorsMinimalChunkQuerySize: Int =
-    Configs.conf.getInt("scheduler.sensors.changedSensorsMinimalChunkQuerySize")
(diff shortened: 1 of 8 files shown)


src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/utilities/Configs.scala
```diff
@@ -83,6 +83,8 @@ object SchedulerConfig {
 object ExecutorsConfig {
   def getThreadPoolSize: Int =
     Configs.conf.getInt("scheduler.executors.thread.pool.size")
+  def getExecutablesFolder: String =
```
kevinwallimann: I think we should use `val`s instead of `def`s in the `Configs.scala`, but especially for this property. If it's not set in the `application.properties`, the shell job will fail silently, because it wraps everything in a `Try`, so the exception here will be swallowed.
In theory, we lose the option to change the property at runtime, but I think we shouldn't do that anyway

jozefbakus: Ok

kevinwallimann (dismissed):

kevinwallimann (dismissed):

kevinwallimann (approved):

PR #397, opened 2021-03-19
Feature/395 prevent self deactivation

### LINKED ISSUE #395: Prevent self-deactivation of scheduler instance
It is possible that an instance deactivates itself (and other instances). This could happen if there's a large delay between `SchedulerInstanceRepository.updateHeartbeat` and `SchedulerInstanceRepository.deactivateLaggingInstance`. An instance should never deactivate itself.

Furthermore, additional log output is needed, if an instance is deactivated

kevinwallimann:
Closes #395


jozefbakus (approved):
Nice

PR #410, opened 2021-03-25
Fix missed notifications in KafkaSensor

### LINKED ISSUE #348: Kafka sensor - missed notifications
After the scheduler restart kafka sensor should pick up all missed messages.
!Be careful about activate/deactive because this should start reading messages from the end of the queue

kevinwallimann:
Closes #348 

**Notes**
The following default values in the `application.properties` were changed:
- Rename `kafkaSource.group.id` to `kafkaSource.group.id.prefix`
  - Changed default value from `hyper_drive_${appUniqueId}` to `hyper_drive`. Unfortunately, substitutions don't seem to work with properties file in typesafe Config. See here for possible explanations: https://gitter.im/lightbend/config?at=57d7edc4b8d4c138408ca85b. Therefore the concatenation is now done in code.
- `kafkaSource.max.poll.records`: Changed default value from 3 to 100. If there are notifications while the app is down, workflows should only be triggered once after the app is restarted. If there are multiple notifications for multiple sensors while the app is down, workflows may be triggered multiple times, if the total number of missed notifications is bigger than `max.poll.records`


src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/sensors/Sensors.scala
```diff
@@ -102,17 +104,18 @@ class Sensors @Inject()(eventProcessor: EventProcessor, sensorRepository: Sensor
     sensors.remove(id)
   }
 
-  private def addNewSensors(assignedWorkflowIds: Seq[Long]): Future[Unit] = {
+  private def addNewSensors(assignedWorkflowIds: Seq[Long], firstIteration: Boolean): Future[Unit] = {
     val activeSensors = sensors.keys.toSeq
     sensorRepository.getNewActiveAssignedSensors(activeSensors, assignedWorkflowIds).map {
-      _.foreach(sensor => startSensor(sensor))
+      _.foreach(sensor => startSensor(sensor, kafkaSensorConsumeFromLatest = !firstIteration))
```
jozefbakus, line 110: It took me a while to understand, but looks good

jozefbakus (approved):
Looks great

PR #432, opened 2021-04-28
Add health indicators

### LINKED ISSUE #425: Health check
Health check rest API implementation.
Health check should include:

- Database connection
- Yarn connection

kevinwallimann:
Closes #425 

There are some built-in health indicators, see https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-features.html#production-ready-health-indicators. DiskSpace Health indicator could be useful and can be enabled in the application properties with `management.health.diskspace.enabled=true`

### DIFF
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -25,8 +25,14 @@ logging.file=${logging.path}/app.log
 management.endpoints.web.base-path=/admin
 management.endpoints.jmx.exposure.exclude=*
 management.endpoints.web.exposure.include=*
+management.endpoint.health.show-details=always
 
 javamelody.management-endpoint-monitoring-enabled=true
+
+# Health check settings
+health.databaseConnection.timeoutMillis=120000
+health.yarnConnection.testEndpoint=/cluster/cluster
+
 # How will users authenticate. Available options: inmemory, ldap
 auth.mechanism=inmemory
 # INMEMORY authentication: username and password defined here will be used for authentication.
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/health/YarnConnectionHealthIndicator.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/health/YarnConnectionHealthIndicator.scala
@@ -1,0 +1,45 @@
+
+/*
+ * Copyright 2018 ABSA Group Limited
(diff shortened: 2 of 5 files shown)


src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/health/YarnConnectionHealthIndicator.scala
```diff
@@ -0,0 +1,45 @@
...
+import za.co.absa.hyperdrive.trigger.scheduler.utilities.{HealthConfig, SparkExecutorConfig}
+
+import java.net.{HttpURLConnection, URL}
+
+@Component
+class YarnConnectionHealthIndicator extends HealthIndicator {
+  val successCode = 200
+
+  override protected def health(): Health = {
+    val yarnBaseUrl = SparkExecutorConfig.getHadoopResourceManagerUrlBase.stripSuffix("/")
```
jozefbakus, line 31: I would wrap every health indicator to try-catch block.
For example my Yarn is down and health check call returns 500 with logged exception: Caused by: java.net.ConnectException: Connection refused (Connection refused)

kevinwallimann: Absolutely. Fixed it

src/main/resources/application.properties
```diff
@@ -25,8 +25,14 @@ logging.file=${logging.path}/app.log
 management.endpoints.web.base-path=/admin
 management.endpoints.jmx.exposure.exclude=*
 management.endpoints.web.exposure.include=*
+management.endpoint.health.show-details=always
 
 javamelody.management-endpoint-monitoring-enabled=true
+
+# Health check settings
+health.databaseConnection.timeoutMillis=120000
```
jozefbakus, line 33: Too long because postgresql will throw exception after 30 seconds (default setting for request timeout)
The same in src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/utilities/Configs.scala

jozefbakus (approved):

src/main/resources/application.properties
jozefbakus: Please ignore. This is true only in case of refused connection

Issue #437, opened 2021-05-03
UX - On an expired session first request is failing


jozefbakus:
UX - On an expired session first request is failing.

This behavior is happening on every page

PR #443, opened 2021-05-11
#437 - UX - On an expired session first rest request is failing

jozefbakus:
Implements: #437 

Changes: 
In case of unauthorized.interceptor.ts -> createRetryAfterAuthenticatedObservable() BaseUrlInterceptor should not be called because request already contains base url.

unauthorized.interceptor.ts should also catch 403 response code.


kevinwallimann (approved):
Awesome! Thanks for fixing this

PR #456, opened 2021-05-24
Pr fix

jozefbakus:
fix for 429 notification crud

Search endpoint is still a bit unstable


kevinwallimann (approved):
What do you mean the search endpoint is unstable?

jozefbakus:
> What do you mean the search endpoint is unstable?

Some of the fields cannot be used for filtering, for example sequences. 
New search filter type must be created.

Issue #446, opened 2021-05-17
Use JsonB type for JobDefinition/JobTemplate properties 


jozefbakus:
Use JsonB type for JobDefinition/JobTemplate properties

PR #459, opened 2021-05-28
#446: Add job type to job instance parameters

jozefbakus:
Implements small part of: #446 

- Need for refactoring because Json format doesn't know how to serialize/deserialize classes like this: 
trait ABTrait
case class A(a: Option[String]) extends ABTrait
case class B(b: Option[String]) extends ABTrait
in case of field a or b is NONE.

### DIFF
--- a/src/main/resources/db_scripts/liquibase/v0.4.3.add-job-type-to-job-instance-parameters.sql
+++ b/src/main/resources/db_scripts/liquibase/v0.4.3.add-job-type-to-job-instance-parameters.sql
@@ -1,0 +1,21 @@
+/*
+ * Copyright 2018 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+update job_instance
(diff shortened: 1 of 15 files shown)


src/main/resources/db_scripts/liquibase/v0.4.3.add-job-type-to-job-instance-parameters.sql
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+update job_instance
+set job_parameters = job_parameters::jsonb || ('{"jobType":"' || job_type || '"}' )::jsonb;
```
jozefbakus, line 17: Could be done also this way: 
```
update job_instance
set job_parameters = jsonb_set(job_parameters, '{jobType}', to_json(job_type::text)::jsonb);
```
but it is not possible because of H2

src/main/scala/za/co/absa/hyperdrive/trigger/models/tables/JobInstanceTable.scala
```diff
@@ -59,21 +56,19 @@ trait JobInstanceTable {
...
-          id = jobInstanceTuple._11
+          jobParameters = jobInstanceTuple._2,
+          jobStatus = jobInstanceTuple._3,
+          executorJobId = jobInstanceTuple._4,
+          applicationId = jobInstanceTuple._5,
+          created = jobInstanceTuple._6,
+          updated = jobInstanceTuple._7,
+          order = jobInstanceTuple._8,
+          dagInstanceId = jobInstanceTuple._9,
+          id = jobInstanceTuple._10
```
kevinwallimann, line 67: Is it possible to use `JobInstance.tupled`?

kevinwallimann (approved):

PR #461, opened 2021-05-31
Feature/455 postgres test

kevinwallimann:


src/test/scala/za/co/absa/hyperdrive/trigger/configuration/liquibase/ApplicationStartPostgresTest.scala
```diff
@@ -43,8 +40,5 @@ class TestApplicationStart extends FlatSpec with Matchers with SpringIntegration
     val workflows = await(workflowRepository.getWorkflows())
     workflows.size shouldBe 1
     workflows.head.name shouldBe workflowJoined.name
-
-    // cleanup
-    h2SchemaDrop()
```
jozefbakus, line 48: I would leave it there.
Currently there is only one test which uses RepositoryPostgresTestBase but in future we will have more.
So clean up is quite fine thing to have

kevinwallimann: It doesn't work, that's why I removed it. The error message is
```
ERROR: "dag_run_view" is not a table
  Hint: Use DROP VIEW to remove a view.
org.postgresql.util.PSQLException: ERROR: "dag_run_view" is not a table
```
Slick expects the dag_run_view to be a table, so it creates it in `dagRunTable.schema.create` as `create table "dag_run_view" ("workflow_id" BIGINT NOT NULL,...` and tries to drop it with `drop table`. That's fine when using `schemaDrop()` together with `schemaSetup()`. However, in this particular test, the schema is created using the liquibase scripts, so dag_run_view is created with `create view`, therefore the above error happens. The ideal thing to do would be to call liquibase to delete the schema, however, I don't expose it in the liquibase config. Another possibility would be to write custom SQL to drop the view.

jozefbakus: Oh, yes. Let's keep it in mind because we will have to fix it in future. 
It is in tests so we can fix it using SQL instead slick table. (replace dagRunTable.schema.drop with sql)

jozefbakus (approved):

PR #462, opened 2021-06-01
Feature/455 notifications sending logic

### LINKED ISSUE #455: Job finish notifications - Backend Sending logic
Given a workflow id `{wid}` a Dag instance status `{status}` and the current timestamp `{curr_time}`, the query to select the matching notification rules at the end of the job looks like that
```
select * from notification_rule nr
join workflow w on w.id = {wid}
left join (select di.workflow_id as wid, extract(epoch from ({curr_time} - di.finished)) as elapsed_secs_last_success from dag_instance di
      where di.status = 'Succeeded'
        and di.workflow_id = {wid}
      order by di.finished desc
      limit 1
) subq on subq.wid = w.id
where w.name like nr.workflow_prefix || '%'
and w.project = (case when length(nr.project) > 0 then nr.project else w.project end)
and nr.statuses ? '{status}'
and (subq.elapsed_secs_last_success is null or subq.elapsed_secs_last_success >= nr.min_elapsed_secs_last_success)
;
```

A notification should be sent
- If the project name matches AND
- If the workflow name matches AND
- If the status matches AND
- If the time since the last successful run is larger than the threshold

Detailed matching logic for each field

- The project name matches either exactly case-insensitively, or if it is not specified (null) it matches everything.
  - E.g. Workflow Project `Project1` matches all rules with project `Project1` or project `<null>`, but not e.g. a rule with project `Project`
- The workflow name matches with the prefix case-insensitively, or if it is not specified (null) it matches everything.
  - E.g. Workflow name `ABC XYZ123 HC` matches a rule with workflow prefix `ABC` or `<null>` or `ABC XYZ`, but not a rule with workflow prefix `XYZ123`
- The status matches if it is equal to any of the statuses of the rule.
  - E.g. If the final status is `Failed`, it matches a rule with statuses `Failed, Succeeded` or `Failed`, but not `<null>`

kevinwallimann:
Closes #455 
**New required application properties**
```
notification.enabled=
notification.sender.address=
spring.mail.host=
spring.mail.port=
```


src/main/resources/logback.xml
```diff
@@ -47,7 +47,8 @@
         <appender-ref ref="FILE-ROLLING"/>
         <appender-ref ref="Console" />
     </logger>
-
+<!--    <logger name="slick.jdbc.JdbcBackend.statement" level="DEBUG" />-->
+<!--    <logger name="slick.jdbc.JdbcBackend.parameter" level="DEBUG" />-->
```
jozefbakus: Why ?

src/main/scala/za/co/absa/hyperdrive/trigger/models/tables/NotificationRuleTable.scala
```diff
@@ -27,31 +28,43 @@ trait NotificationRuleTable extends SearchableTableQuery {
   import api._
 
   final class NotificationRuleTable(tag: Tag) extends Table[NotificationRule](tag, _tableName = "notification_rule") with SearchableTable {
+    def isActive: Rep[Boolean] = column[Boolean]("is_active")
+
```
jozefbakus: Could you please remove empty lines here ?

src/main/scala/za/co/absa/hyperdrive/trigger/models/tables/NotificationRuleTable.scala
```diff
@@ -27,31 +28,43 @@ trait NotificationRuleTable extends SearchableTableQuery {
   import api._
 
   final class NotificationRuleTable(tag: Tag) extends Table[NotificationRule](tag, _tableName = "notification_rule") with SearchableTable {
```
jozefbakus, line 30: Missing isActive delta script

src/main/resources/logback.xml
kevinwallimann: Sorry, forgot to remove

src/main/scala/za/co/absa/hyperdrive/trigger/models/tables/NotificationRuleTable.scala
kevinwallimann: Ok, for some reason, that's the IntelliJ autoformat

src/main/scala/za/co/absa/hyperdrive/trigger/models/tables/NotificationRuleTable.scala
kevinwallimann: Fixed

jozefbakus (approved):
Cool, works fine

PR #474, opened 2021-06-23
Add timeout config for yarn connection health indicator

### LINKED ISSUE #470: Add timeout for YarnConnectionHealthIndicator
If the Yarn Url is unreachable, the /admin/health endpoint will suffer from a timeout. A configurable timeout should be added for the connection. `URLConnection.setConnectTimeout` can be used to implement it

**New application property**
```
health.yarnConnection.timeoutMillis
```
Default value = 60000

kevinwallimann:
Closes #470 

**New application property**
```
health.yarnConnection.timeoutMillis
```
Default value = 60000

### DIFF
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -30,7 +30,7 @@ management.endpoint.health.show-details=always
 javamelody.management-endpoint-monitoring-enabled=true
 
 # Health check settings
-health.databaseConnection.timeoutMillis=120000
+health.databaseConnection.timeoutMillis=60000
 health.yarnConnection.testEndpoint=/cluster/cluster
 health.yarnConnection.timeoutMillis=60000
 
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -32,6 +32,7 @@ javamelody.management-endpoint-monitoring-enabled=true
 # Health check settings
 health.databaseConnection.timeoutMillis=120000
 health.yarnConnection.testEndpoint=/cluster/cluster
+health.yarnConnection.timeoutMillis=120000
 
 # How will users authenticate. Available options: inmemory, ldap
 auth.mechanism=inmemory
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -32,7 +32,7 @@ javamelody.management-endpoint-monitoring-enabled=true
 # Health check settings
 health.databaseConnection.timeoutMillis=120000
 health.yarnConnection.testEndpoint=/cluster/cluster
-health.yarnConnection.timeoutMillis=120000
+health.yarnConnection.timeoutMillis=60000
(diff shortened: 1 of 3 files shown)


src/main/resources/application.properties
```diff
@@ -32,6 +32,7 @@ javamelody.management-endpoint-monitoring-enabled=true
 # Health check settings
 health.databaseConnection.timeoutMillis=120000
 health.yarnConnection.testEndpoint=/cluster/cluster
+health.yarnConnection.timeoutMillis=120000
```
jozefbakus: Could you please add mention of this prop to the PR description? 
When I do release notes it is easier to scrap changes

src/main/resources/application.properties
```diff
@@ -32,6 +32,7 @@ javamelody.management-endpoint-monitoring-enabled=true
 # Health check settings
 health.databaseConnection.timeoutMillis=120000
```
jozefbakus: I would decrease timeouts to a max of 1 minute.

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/utilities/Configs.scala
```diff
@@ -122,8 +122,11 @@ object JobDefinitionConfig {
 object HealthConfig {
   lazy val databaseConnectionTimeoutMillis: Int =
     Try(Configs.conf.getInt("health.databaseConnection.timeoutMillis")).getOrElse(120000)
-  lazy val yarnConnectionTestEndpoint: String =
+  lazy val yarnConnectionTestEndpoint: String = {
     Configs.conf.getString("health.yarnConnection.testEndpoint")
+  }
+  lazy val yarnConnectionTimeoutMillisOpt: Option[Int] =
```
jozefbakus, line 128: What about no option with default value like in databaseConnectionTimeoutMillis

kevinwallimann: You don't have to set a connection timeout on the `URLConnection`. If I provide a default value, you cannot configure that behaviour anymore.

src/main/resources/application.properties
kevinwallimann: Done

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/utilities/Configs.scala
jozefbakus: But why the different behavior for yarn compared to DB?

src/main/resources/application.properties
kevinwallimann: Done

jozefbakus (approved):

PR #475, opened 2021-06-24
#466: Kill running Spark Job

### LINKED ISSUE #466: Kill running Spark Job
Once spark job is submitted
User should be able to see kill job button on runs detail page and
should be able to kill running spark job

On the backed side we should use yarn application id and YARN REST API: 
http://hadoop.apache.org/docs/r2.5.2/hadoop-yarn/hadoop-yarn-site/ResourceManagerRest.html#Cluster_Application_State_API

jozefbakus:
Implements: #466 - Kill running Spark Job

User is only able to kill Running Spark job with application id from yarn

New application property:
sparkYarnSink.userUsedToKillJob

### DIFF
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/JobInstanceController.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/JobInstanceController.scala
@@ -32,8 +32,8 @@ class JobInstanceController @Inject()(jobInstanceService: JobInstanceService) {
     jobInstanceService.getJobInstances(dagInstanceId).toJava.toCompletableFuture
   }
 
-  @PostMapping(path = Array("/jobInstances"))
-  def killJob(@RequestParam applicationId: String): CompletableFuture[Boolean] = {
+  @PostMapping(path = Array("/jobInstances/{applicationId}/kill"))
+  def killJob(@PathVariable applicationId: String): CompletableFuture[Boolean] = {
     jobInstanceService.killJob(applicationId).toJava.toCompletableFuture
   }
 
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/JobInstanceController.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/JobInstanceController.scala
@@ -16,7 +16,6 @@
 package za.co.absa.hyperdrive.trigger.api.rest.controllers
 
 import java.util.concurrent.CompletableFuture
-
 import za.co.absa.hyperdrive.trigger.api.rest.services.JobInstanceService
 import za.co.absa.hyperdrive.trigger.models.JobInstance
 import javax.inject.Inject
@@ -31,6 +30,11 @@ class JobInstanceController @Inject()(jobInstanceService: JobInstanceService) {
   @GetMapping(path = Array("/jobInstances"))
   def getJobInstances(@RequestParam dagInstanceId: Long): CompletableFuture[Seq[JobInstance]] = {
     jobInstanceService.getJobInstances(dagInstanceId).toJava.toCompletableFuture
+  }
+
+  @PostMapping(path = Array("/jobInstances"))
+  def killJob(@RequestParam applicationId: String): CompletableFuture[Boolean] = {
+    jobInstanceService.killJob(applicationId).toJava.toCompletableFuture
   }
 
 }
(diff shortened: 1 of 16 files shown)


src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/JobInstanceController.scala
```diff
@@ -33,5 +32,10 @@ class JobInstanceController @Inject()(jobInstanceService: JobInstanceService) {
     jobInstanceService.getJobInstances(dagInstanceId).toJava.toCompletableFuture
   }
 
+  @PostMapping(path = Array("/jobInstances"))
```
kevinwallimann: We need to clarify anyway how we want to deal with custom actions (search, activate, kill), but for the time being, it's pretty surprising that POST /jobInstances leads to killing 1 job. 1. It should not be plural, 2. POST /entityName usually maps to creating the object. I'd use `/jobInstance/kill` for the moment

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/JobInstanceService.scala
```diff
@@ -33,4 +39,19 @@ class JobInstanceServiceImpl(override val jobInstanceRepository: JobInstanceRepo
...
+    val url: String =
+      s"${SparkExecutorConfig.getHadoopResourceManagerUrlBase}/ws/v1/cluster/apps/$applicationId/state?user.name=${SparkExecutorConfig.getUserUsedToKillJob}"
+    val data = Json.obj(
+      "state" -> "KILLED"
+    )
+
+    Try {
+      WSClientProvider.getWSClient.url(url).put(data).map { response =>
+        response.status == 200 || response.status == 202
+      }.recoverWith {
```
kevinwallimann, line 52: Nice, so there's no need to update any status, because the status is updated anyway by the usual scheduler loop. That's really nice and decoupled

ui/src/app/constants/texts.constants.ts
```diff
@@ -51,6 +51,12 @@ export const texts = {
   RUN_WORKFLOWS_SUCCESS_NOTIFICATION: 'Workflows have been submitted.',
   RUN_WORKFLOWS_FAILURE_NOTIFICATION: "Sorry, workflows couldn't be submitted. Please try again.",
 
+  KILL_JOB_CONFIRMATION_TITLE: 'Kill running job',
+  KILL_JOB_CONFIRMATION_CONTENT:
+    'Are you sure you want to kill running job? The operation cannot be reverted and operation may not take effect.',
```
kevinwallimann: This is correct, but it sounds a bit confusing. I would only keep "The operation cannot be reverted" and omit "operation may not take effect"

ui/src/app/constants/texts.constants.ts
```diff
@@ -51,6 +51,12 @@ export const texts = {
...
+  KILL_JOB_SUCCESS_NOTIFICATION: 'Kill job request passed, but it does not mean that job is killed. Please refresh page and verify status.',
```
kevinwallimann: I'd rephrase "Request to kill job passed" Otherwise it is a bit confusing it could mean killing a job request (whatever a job request is) or requesting to kill the job. Again it is 100% correct, but a bit confusing. How about

"Request to kill the job passed. Please refresh the page after a few seconds and verify its status"

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/JobInstanceService.scala
jozefbakus: Exactly

ui/src/app/constants/texts.constants.ts
jozefbakus: Done

ui/src/app/constants/texts.constants.ts
jozefbakus: Ok

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/JobInstanceController.scala
jozefbakus: I will follow the current convention and we can start thinking about REST API refactoring and follow rest paradigm

kevinwallimann (approved):

PR #481, opened 2021-07-08
Add logging for failed futures

### LINKED ISSUE #473: Add error handling for futures
Currently, no exception is logged if one of the futures fail here: https://github.com/AbsaOSS/hyperdrive-trigger/blob/develop/src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/Executors.scala#L47-L57

Since the outcome is also not logged in `JobScheduler.scala`, any exceptions thrown in these futures are silently swallowed. This is unintended. Exceptions in the futures should be logged with level Error using `.onComplete`. Successful futures may or may not be logged (open for discussion)

`.onComplete` error handling should also be added for the `runningAssignWorkflows` future in `JobScheduler.scala`

kevinwallimann:
Closes #473

### DIFF
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/Executors.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/Executors.scala
@@ -44,17 +44,27 @@ class Executors @Inject()(dagInstanceRepository: DagInstanceRepository, jobInsta
     jobInstanceRepository.getJobInstances(dagInstance.id).flatMap {
       case jobInstances if jobInstances.exists(_.jobStatus.isFailed) =>
         val updatedDagInstance = dagInstance.copy(status = DagInstanceStatuses.Failed, finished = Option(LocalDateTime.now()))
-        for {
+        val fut = for {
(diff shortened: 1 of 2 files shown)


src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/Executors.scala
```diff
@@ -44,17 +44,27 @@ class Executors @Inject()(dagInstanceRepository: DagInstanceRepository, jobInsta
...
         } yield {}
+        fut.onComplete {
+          case Failure(exception) => logger.error(s"Updating status failed for failed run. Dag instance id = ${dagInstance.id}", exception)
+          case _ =>
+        }
+        fut
       case jobInstances if jobInstances.forall(ji => ji.jobStatus.isFinalStatus && !ji.jobStatus.isFailed) =>
         val updatedDagInstance = dagInstance.copy(status = DagInstanceStatuses.Succeeded, finished = Option(LocalDateTime.now()))
-        for {
+        val fut = for {
```
jozefbakus, line 59: I think it should be enough to have it logged at the end and not in every case.

jozefbakus (approved):

PR #485, opened 2021-07-12
Immediately fail when application couldn't be submitted and was kille…

### LINKED ISSUE #457: Job status should be updated immediately if the submission failed
If the job submission fails, e.g. because a file could not be found, the status is only updated after the submission timeout. Probably this can be easily changed by changing this line in SparkExecutor

Current
```
        override def stateChanged(handle: SparkAppHandle): Unit =
          if (handle.getState == SparkAppHandle.State.SUBMITTED) {
            latch.countDown()
          }
```

Proposed change
```
        override def stateChanged(handle: SparkAppHandle): Unit =
          if (handle.getState >= SparkAppHandle.State.SUBMITTED) {
            latch.countDown()
          }
```
This would include the states SUBMITTED, RUNNING, FINISHED, FAILED and KILLED.

kevinwallimann:
…d directly
Closes #457


src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/spark/SparkExecutor.scala
```diff
@@ -50,8 +50,9 @@ object SparkExecutor extends Executor[SparkInstanceParameters] {
       val submitTimeOut = SparkExecutorConfig.getSubmitTimeOut
       val latch = new CountDownLatch(1)
       val sparkAppHandle = getSparkLauncher(id, ji.jobName, jobParameters).startApplication(new SparkAppHandle.Listener {
+        import scala.math.Ordered.orderingToOrdered
         override def stateChanged(handle: SparkAppHandle): Unit =
-          if (handle.getState == SparkAppHandle.State.SUBMITTED) {
+          if (handle.getState >= SparkAppHandle.State.SUBMITTED) {
```
jozefbakus, line 55: How do you know the order of statuses? 
SparkAppHandle.State does not implement Ordered interface.

jozefbakus (approved):

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/spark/SparkExecutor.scala
jozefbakus: I tested it and it works as expected

PR #493, opened 2021-07-27
Feature/467 poc emr submitter

### LINKED ISSUE #467: Submit jobs to EMR with AWS SDK
The application should be able to submit spark job to AWS-EMR using AWS-steps instead of SparkLauncher and Yarn

kevinwallimann:
Closes #467 

- `spark.submitApi`. Must be either `yarn` or `emr`. Default value: `yarn`
- `spark.emr.clusterId`. The Id of the EMR-cluster (e.g. j-2AXXXXXXGAPLF) Mandatory, if `spark.submitApi=emr`
- `spark.emr.filesToDeploy`. Optional
- `spark.emr.additionalConfs`. Optional. This is just a prefix and works like the counterpart `sparkYarnSink.additionalConfs`
- `spark.emr.awsProfile`. Optional. Intended for local development
- `spark.emr.region`. Optional. Intended for local development

The idea is that in production, AWS credentials are obtained through a role

**sparkYarnSink properties**
If `spark.submitApi` is `emr`, the following properties will not be used
- `sparkYarnSink.hadoopConfDir`
- `sparkYarnSink.sparkHome`
- `sparkYarnSink.master`
- `sparkYarnSink.submitTimeout`
- `sparkYarnSink.filesToDeploy`
- `sparkYarnSink.additionalConfs` (all properties prefixed with this)
- `sparkYarnSink.executablesFolder`

Note that 
- `sparkYarnSink.hadoopResourceManagerUrlBase`
- `sparkYarnSink.userUsedToKillJob`
are still required even if `spark.submitApi=emr`, because only the job submission is done with the AWS SDK (for now), but not the status updates, killing job etc.


src/main/resources/application.properties
```diff
@@ -106,7 +108,12 @@ sparkYarnSink.additionalConfs.spark.yarn.principal=
 sparkYarnSink.additionalConfs.spark.shuffle.service.enabled=true
 sparkYarnSink.additionalConfs.spark.dynamicAllocation.enabled=true
 sparkYarnSink.executablesFolder=/
-sparkYarnSink.userUsedToKillJob=
+
+spark.emr.clusterId=
+spark.emr.filesToDeploy=
+spark.emr.additionalConfs=
+spark.emr.awsProfile=
+spark.emr.region=
```
Zejnilovic, line 116: So our EMR clusters will need to be shut down each ... let's say month as they are not built for long term use. Do you use any properties hot-swapping or would this need to be "redeployed" every time we have to redeploy EMR?

kevinwallimann: Currently, there is no such mechanism. One way would be to move the clusterId to the workflow config, i.e. the config value would end up in the database and can be updated without an application restart. Otherwise, we'd have to look into what spring offers in terms of properties hot-swapping, esp. if that would be relevant for other applications as well, especially on cloud

Zejnilovic: > if that would be relevant for other applications as well, especially on cloud

I think it would be. But I do not want to burden you with it right now. I can research myself a bit as well.

All in all a good PR. Please take this as just passing thoughts that is good to keep in mind.

jozefbakus (approved):
Looks good !

PR #495, opened 2021-07-29
#489, #490, #494  Config fix

### LINKED ISSUE #489: Use auth inmemory config properties
Currently, the in-memory username and password are hard-coded. The values should be set by the config properties `auth.inmemory.user` and `auth.inmemory.password` or these config properties should be removed from the `application.properties`.

### LINKED ISSUE #490: Use the shellExecutor.executablesFolder property
Currently the `shellExecutor.executablesFolder` property is only used in a test. It should be used in the `ShellExecutor` and prepend paths to shell scripts. If it doesn't need to be used, the property should be removed.

### LINKED ISSUE #494: Use the sparkYarnSinkConfig.executablesFolder property
Currently the sparkYarnSinkConfig.executablesFolder property is only used in a test. It should be used in the SparkExecutor and prepend paths to files. If it doesn't need to be used, the property should be removed.

jozefbakus:
Implements: #489 - Use auth inmemory config properties
Implements: #490 - Use the shellExecutor.executablesFolder property
Implements: #494 - Use the sparkYarnSinkConfig.executablesFolder property

### DIFF
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/JobTemplateService.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/JobTemplateService.scala
@@ -16,7 +16,6 @@
 package za.co.absa.hyperdrive.trigger.api.rest.services
 
 import org.springframework.stereotype.Service
-import za.co.absa.hyperdrive.trigger.api.rest.utils.JobTemplateResolutionUtil
 import za.co.absa.hyperdrive.trigger.models.search.{TableSearchRequest, TableSearchResponse}
 import za.co.absa.hyperdrive.trigger.models.{DagDefinitionJoined, JobTemplate, ResolvedJobDefinition}
 import za.co.absa.hyperdrive.trigger.persistance.JobTemplateRepository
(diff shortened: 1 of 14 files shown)


kevinwallimann (dismissed):

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/utils/JobTemplateResolutionUtil.scala
```diff
@@ -15,13 +15,17 @@
...
+import za.co.absa.hyperdrive.trigger.configuration.application.{ShellExecutorConfig, SparkYarnSinkConfig}
 
+import java.nio.file.Paths
+import javax.inject.Inject
 import scala.util.{Failure, Success, Try}
 
-
-object JobTemplateResolutionUtil {
+@Component
+class JobTemplateResolutionUtil @Inject()(sparkYarnSinkConfig: SparkYarnSinkConfig, shellExecutorConfig: ShellExecutorConfig) {
```
kevinwallimann: I think you should rename the class so it doesn't end in `Util` and also move it to some other package, since utils are supposed to be singleton objects. Otherwise, you can pass the executables folder variables.

jozefbakus: Ok, I made service of it

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/JobTemplateService.scala
```diff
@@ -35,15 +34,15 @@ trait JobTemplateService {
 }
 
 @Service
-class JobTemplateServiceImpl(override val jobTemplateRepository: JobTemplateRepository) extends JobTemplateService {
+class JobTemplateServiceImpl(override val jobTemplateRepository: JobTemplateRepository, jobTemplateResolutionUtil: JobTemplateResolutionService) extends JobTemplateService {
```
kevinwallimann: rename the variable to service as well

jozefbakus: Fixed

kevinwallimann (approved):

PR #501, opened 2021-08-26
Fix db_script_latest.sql

### LINKED ISSUE #499: Fix db_scripts_latest.sql
Since the db_scripts_latest.sql is not routinely used, a few errors went unnoticed.

- Order of inserted fields (may be important to have the same order of fields on a table whether they were generated by liquibase or by applying the insert script. Important e.g. when inserting data using copy statements)
- Insert into statements (insert into job_template)

kevinwallimann:
Closes #499


jozefbakus (approved):
I've checked also the last few delta scripts changes and it looks up to date.
+ I compared tables generated by Liquidbase and latest script.

PR #505, opened 2021-08-31
Fix compilation

DKTRZHR000:


kevinwallimann (approved):

kevinwallimann:
Thanks a lot @DoktorZhor I'm not quite sure why it has worked previously with other annotations. Maybe it's because I'm using it on the class level

PR #511, opened 2021-09-06
Slashes in jar paths should not be removed

### LINKED ISSUE #509: Slashes in jar paths should not be removed
Most probably due to #494 and #490, the following test case in `JobTemplateResolutionServiceTest` fails.
```
  it should "correctly apply the executables folder" in {
    // given
    val jobTemplate = GenericSparkJobTemplate
    val jobJar = "s3://abc/xyz/ghi-123.jar"
    val jobDefinition = createJobDefinition().copy(jobTemplateId = jobTemplate.id, jobParameters = SparkDefinitionParameters(
      jobJar = Some(jobJar), mainClass = None
    ))
    val dagDefinitionJoined = createDagDefinitionJoined(jobDefinition)

    // when
    val resolvedJobDefinitions = underTest.resolveDagDefinitionJoined(dagDefinitionJoined, Seq(jobTemplate))

    // then
    val resolvedJobDefinition = resolvedJobDefinitions.head
    resolvedJobDefinition.jobParameters.asInstanceOf[SparkInstanceParameters].jobJar shouldBe jobJar
  }
```

The test fails with
"s3:/[]abc/xyz/ghi-123.jar" was not equal to "s3:/[/]abc/xyz/ghi-123.jar"
Instead of retaining the two slashes after the colon, one slash was removed. It means that for any workflow that references a jar starting with `hdfs://` or `s3://`, the path of the jar is changed to `hdfs:/` and `s3:/`, which will most probably lead to the failure of the spark job. This happens even if `sparkYarnSink.executablesFolder` is not set, so there is no workaround.

As with the original issues #494 and #490, I suggest either fixing the problem or removing the feature altogether, since it isn't being used.

jozefbakus:
Closes: #509 

**Application properties changes**
Removed
- `shellExecutor.executablesFolder` 
- `sparkYarnSink.executablesFolder`


src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/utils/PathResolverUtil.scala
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package za.co.absa.hyperdrive.trigger.api.rest.utils
+
+object PathResolverUtil {
+  def resolve(basePath: String, childPath: String): String = {
+    basePath match {
+      case bp: String if bp.isEmpty => childPath
+      case bp: String if childPath.startsWith("/") => s"${bp}${childPath.replaceFirst("/", "")}"
```
kevinwallimann: Does this make sense? If `basePath` is `/tmp` and ` childPath` is `/some/path` then the result would be `/tmpsome/path`

kevinwallimann (approved):

PR #515, opened 2021-09-13
Don't update job status to submissionTimeout for EMR

### LINKED ISSUE #513: JobStatus should not go to SubmissionTimeout while job is pending when using AddJobFlowStep
**Description**
Currently, when using `spark.submitApi=emr`, a job will always go to jobStatus `SubmissionTimeout`. That's because `SparkEmrClusterServiceImpl.submitJob` returns immediately. However, it takes a certain amount of time, until the job appears in the YARN ResourceManager. During this time, `SparkExecutor.updateJobStatus` will set the jobStatus to `SubmissionTimeout`.

**Solution**
The logic in https://github.com/AbsaOSS/hyperdrive-trigger/blob/develop/src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/spark/SparkExecutor.scala#L48
could be moved to the `SparkClusterService` trait, with specific implementations for `SparkYarnClusterServiceImpl` and `SparkEmrClusterServiceImpl`.

In the future, the whole job status update logic could be specialized for Emr, with #497

kevinwallimann:
Closes #513


src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/spark/SparkEmrClusterServiceImpl.scala
```diff
@@ -58,6 +58,11 @@ class SparkEmrClusterServiceImpl @Inject()(sparkConfig: SparkConfig, emrClusterP
     }
   }
 
+  override def handleMissingYarnStatusForJobStatusSubmitting(jobInstance: JobInstance, updateJob: JobInstance => Future[Unit])
+    (implicit executionContext: ExecutionContext): Future[Unit] = Future {
+    // do nothing
```
jozefbakus: If something goes wrong, job will be never executed and stay in submitting status.

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/spark/SparkEmrClusterServiceImpl.scala
```diff
@@ -55,12 +57,28 @@ class SparkEmrClusterServiceImpl @Inject()(sparkConfig: SparkConfig, emrClusterP
       val stepId = response.getStepIds.asScala.headOption
       logger.info(s"Added jobFlowStepsRequest ${jobFlowStepsRequest} for executorId ${id} and stepId $stepId}")
       logger.info(response.toString)
+      stepId
+    }.flatMap { stepId =>
+      val jiStepId = jiSubmitting.copy(stepId = stepId)
+      updateJob(jiStepId)
```
jozefbakus, line 66: If updateJob doesn't pass it will end up with status Lost and will get executed again (2 instances of the same workflow/job).
Is it possible to pass unique identifier into step? Similar to yarn?

kevinwallimann: @jozefbakus I updated the PR, as discussed

jozefbakus (approved):

Issue #521, opened 2021-10-18
Hyperdrive job ui component - dedicated ui for Hyperdrive job types


jozefbakus:
Fields of the form component should be shown based on the selected template:

No template: 
- Component should behave like a spark job.

Selected template: 
- Fields should be derived based on the selected template
-- Offload Raw Topic with HyperConformance
-- Offload Publish Topic
-- HyperConformance Raw to Publish Topic
- In every case let user set also Additional jars, Additional files, Additional Spark Config and App arguments

PR #529, opened 2021-11-04
Feature/521 hyperdrive job UI component

jozefbakus:
Implements: #521 

Implemented hyperdrive ingestion types: 
- HyperConformance Raw to Publish Topic
- Offload Raw Topic with HyperConformance
- Offload Publish Topic


ui/src/app/components/workflows/workflow-form/jobs/job/job-types/hyperdrive-job/hyperdrive-job.component.ts
```diff
@@ -27,20 +37,81 @@ export class HyperdriveJobComponent implements OnInit, OnDestroy {
...
+  }
+
+  getHyperdriveType(): string {
+    return this.hyperdriveType;
+  }
+
+  hyperdriveTypeChange(value: string) {
+    this.jobParametersChange.emit({
+      ...this.jobParameters,
+      jobJar: this.jobParameters.jobJar,
```
kevinwallimann: Sometimes you use the elvis operator for `this.jobParameters`, but sometimes (e.g. here) you don't. It should be consistent imo. If you look at this component in isolation, you should have the elvis operator everywhere.

ui/src/app/components/workflows/workflow-form/jobs/job/job-types/hyperdrive-job/hyperdrive-job.component.ts
```diff
@@ -27,20 +37,81 @@ export class HyperdriveJobComponent implements OnInit, OnDestroy {
...
+      mainClass: this.jobParameters.mainClass,
+      appArguments: this.getAppArguments(),
+    });
+    this.hyperdriveType = value;
+  }
+
+  getHyperdriveFieldValue(fieldPrefix: string): any {
+    return this.jobParameters?.appArguments.find((appArgument) => appArgument.startsWith(fieldPrefix))?.replace(fieldPrefix, '');
```
kevinwallimann, line 102: Btw I think internally representing the `appArguments` as a Map, split by `=` will clean up a lot of this code with `startsWith` etc. I think you could even attempt it as part of this PR, as an implementation detail (without UI changes)

ui/src/app/components/workflows/workflow-form/jobs/job/job-types/hyperdrive-job/hyperdrive-job.component.ts
```diff
@@ -27,20 +37,81 @@ export class HyperdriveJobComponent implements OnInit, OnDestroy {
...
+      if (value?.jobTemplateId) {
+        const jobTemplateParameters = <HyperdriveTemplateParametersModel>value?.jobTemplateParameters;
+        this.hyperdriveType = HyperdriveUtil.getHyperdriveTypeFromAppArguments(
+          jobTemplateParameters.appArguments,
+          hyperdriveTypesJobTemplateFields,
+        );
       }
     });
+
+    if (this.isJobTemplateSelected) {
```
kevinwallimann, line 80: I would extract this block to a separate function `initHyperdriveType` or something like that

ui/src/app/components/workflows/workflow-form/jobs/job/job-types/hyperdrive-job/hyperdrive-job.component.ts
```diff
@@ -27,20 +37,81 @@ export class HyperdriveJobComponent implements OnInit, OnDestroy {
...
+      const hyperdriveTypeFromJobParams = HyperdriveUtil.getHyperdriveTypeFromAppArguments(
+        this.jobParameters.appArguments,
+        hyperdriveTypesFields,
+      );
+      const hyperdriveTypeFromTemplateParams = HyperdriveUtil.getHyperdriveTypeFromAppArguments(
+        this.jobTemplateParameters.appArguments,
+        hyperdriveTypesJobTemplateFields,
+      );
+
+      if (hyperdriveTypeFromJobParams == hyperdriveTypeFromTemplateParams) {
```
kevinwallimann: Is this really necessary to check the hyperdrive type from `jobParameters.appArguments`? If yes, why do you not check it in the `jobTemplateChangesSubscription`? Shouldn't the type derived from the template take precedence if you really had a job where the hyperdrive type conflicts for whatever reason?

jozefbakus: I think it is necessary to check the hyperdrive type, because for example if you are in show mode and the template would take precedence hyperdrive type would not be correct. I decided to show hyperdrive type only if it is 100% sure.

ui/src/app/components/workflows/workflow-form/jobs/job/job-types/hyperdrive-job/hyperdrive-job.component.ts
jozefbakus: I would leave it and solve it once by splitting it to key value in backend as well. I am not really sure how to do it.

ui/src/app/components/workflows/workflow-form/jobs/job/job-types/hyperdrive-job/hyperdrive-job.component.ts
jozefbakus: Fixed

ui/src/app/components/workflows/workflow-form/jobs/job/job-types/hyperdrive-job/hyperdrive-job.component.ts
jozefbakus: Done

ui/src/app/components/workflows/workflow-form/jobs/job/job-types/hyperdrive-job/hyperdrive-job.component.ts
```diff
@@ -68,10 +68,13 @@ export class HyperdriveJobComponent implements OnInit, OnDestroy {
         );
       }
     });
+    this.initHyperdriveType();
+  }
 
+  initHyperdriveType() {
```
kevinwallimann: Actually, I think you could even make it better by returning the type and making this a function without side effects.

ui/src/app/components/workflows/workflow-form/jobs/job/job-types/hyperdrive-job/hyperdrive-job.component.ts
kevinwallimann: Ok, but then I still think that the logic should be the same in the jobTemplateSubscription. You might also want to consider moving this logic to the util

jozefbakus: It is not really the same logic, it has to be executed also outside of the jobTemplateSubscription because logic inside is executed only on job template change so it would not get executed on init. 
If you remove job template you don't have to run the logic of the finding correct hyperdrive type.

Moved one part into util

ui/src/app/components/workflows/workflow-form/jobs/job/job-types/hyperdrive-job/hyperdrive-job.component.ts
jozefbakus: Fixed

ui/src/app/components/workflows/workflow-form/jobs/job/job-types/hyperdrive-job/hyperdrive-job.component.ts
```diff
@@ -27,20 +37,76 @@ export class HyperdriveJobComponent implements OnInit, OnDestroy {
...
   ngOnInit(): void {
-    this.jobTemplateChangesSubscription = this.jobTemplateChanges.subscribe((value) => {
-      if (value) {
-        this.jobParametersChange.emit({ ...this.jobParameters, jobJar: undefined, mainClass: undefined });
+    this.jobTemplateChangesSubscription = this.jobTemplateChanges.subscribe((value: JobTemplateChangeEventModel) => {
+      this.jobParametersChange.emit({
+        ...this.jobParameters,
+        jobJar: undefined,
+        mainClass: undefined,
+        appArguments: this.getAppArguments(),
```
kevinwallimann: If you change the job template, then the source topic is lost. I think it's because you filter the job template arguments out here.

jozefbakus: I made it simpler, I just clean all app arguments on hyperdrive type or job template change.
If we decide to keep app arguments we can make it later

ui/src/app/components/workflows/workflow-form/jobs/job/job-types/hyperdrive-job/hyperdrive-job.component.ts
jozefbakus: After discussion, I've refactored it

ui/src/app/components/workflows/workflow-form/jobs/job/job-types/hyperdrive-job/hyperdrive-job.component.ts
kevinwallimann: I should not have said anything, this was one of my favorite features 😅

kevinwallimann (dismissed):

ui/src/app/components/workflows/workflow-form/jobs/job/job-types/hyperdrive-job/hyperdrive-job.component.ts
jozefbakus: Oh :/ It's back

kevinwallimann (approved):

PR #531, opened 2021-11-08
517 Menas integration - search dataset endpoint

### LINKED ISSUE #517: Integration with Menas - Search datasets endpoint
Implement Search dataset endpoint (Calling Menas rest API):
- Menas search endpoint should return a list of pairs (dataset name, latest dataset version)
- Use /api/dataset/list/{searchQuery} https://github.com/AbsaOSS/enceladus/blob/eecab712b22b4e1b5cd75fb1376d3c6d2d4b8405/menas/src/main/scala/za/co/absa/enceladus/menas/controllers/VersionedModelController.scala#L39
- Endpoint should return list of pairs (datasetName, latestVersion) and should be able to pass searchQuery
- Authentication with menas should be implemented. For more info contact (@kevinwallimann, @jozefbakus or @Zejnilovic)

DKTRZHR000:
Closes #517


src/main/resources/application.properties
```diff
@@ -120,3 +120,11 @@ db.numThreads=4
 
 db.skip.liquibase=false
 spring.liquibase.change-log=classpath:/db_scripts/liquibase/db.changelog.yml
+
+menas.keytab=
```
jozefbakus: I would do it here similar way like for auth.mechanism=<auth mechanism (in-memory/LDAP)>

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/client/AuthClient.scala
```diff
@@ -0,0 +1,107 @@
...
+
+class SpnegoAuthClient(username: String, keytabLocation: String, restTemplate: RestTemplate, apiCaller: ApiCaller, path: String)
+    extends AuthClient(username, restTemplate, apiCaller, baseUrl => s"$baseUrl$path") {
+  override protected def requestAuthentication(url: String): ResponseEntity[String] = {
+    logger.info(s"Authenticating via SPNEGO ($url): user `$username`, with keytab `$keytabLocation`")
+    restTemplate.getForEntity(url, classOf[String])
+  }
+}
+
+class LdapAuthClient(username: String, password: String, restTemplate: RestTemplate, apiCaller: ApiCaller, path: String)
```
jozefbakus: I think it is not really a Ldap. Menas is used to auth. What they use is irrelevant I think

DKTRZHR000: I see. Do we have some good name for it? All that comes to mind for username/password authentication is the word `standard` ^^

jozefbakus: No Problem with standard :D

jozefbakus (commented):
In general, it looks awesome.
Is it possible to separate Rest client, ApiCaller and rest from Menas part ? 
For example moving menas part into services package. 
My point is that in near future I will need to reuse the rest client.

kevinwallimann:
Closed because outdated and won't be fixed

PR #533, opened 2021-11-18
#523: Check if kafka topic exist

### LINKED ISSUE #523: Smart UI - verify if kafka topic exists
Implement endpoint which verifies if kafka topic exists:

jozefbakus:
Implements #523 

- Exposed new endpoint to check if kafka topic exist 
- Endpoint: GET /doesKafkaTopicExist?kafkaTopicName=<topic-name>&servers=<server1>,<server2>

New application property:
`kafkaSource.connectionTimeoutMillis`

### DIFF
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -69,6 +69,7 @@ spring.mail.port=
 #Kafka sensor properties.
 kafkaSource.group.id.prefix=hyper_drive
 kafkaSource.poll.duration=500
+kafkaSource.connectionTimeoutMillis=10000
 kafkaSource.properties.key.deserializer=org.apache.kafka.common.serialization.StringDeserializer
 kafkaSource.properties.value.deserializer=org.apache.kafka.common.serialization.StringDeserializer
 kafkaSource.properties.max.poll.records=100
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/KafkaController.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/KafkaController.scala
@@ -24,8 +24,8 @@ import scala.concurrent.ExecutionContext.Implicits.global
 
 @RestController
 class KafkaController @Inject()(kafkaService: KafkaService) {
-  @GetMapping(path = Array("/doesKafkaTopicExist"))
-  def doesKafkaTopicExist(@RequestParam kafkaTopicName: String, @RequestParam servers: Array[String]): CompletableFuture[Boolean] = {
-    kafkaService.doesKafkaTopicExist(kafkaTopicName, servers).toJava.toCompletableFuture
+  @GetMapping(path = Array("/topics/{topicName}/exists"))
+  def existsTopic(@PathVariable topicName: String, @RequestParam servers: Array[String]): CompletableFuture[Boolean] = {
+    kafkaService.existsTopic(topicName, servers).toJava.toCompletableFuture
   }
 }
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/KafkaController.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/KafkaController.scala
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2018 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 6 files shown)


src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/KafkaController.scala
```diff
@@ -0,0 +1,31 @@
...
+import org.springframework.web.bind.annotation._
+import za.co.absa.hyperdrive.trigger.api.rest.services.KafkaService
+import java.util.concurrent.CompletableFuture
+import javax.inject.Inject
+import scala.compat.java8.FutureConverters._
+import scala.concurrent.ExecutionContext.Implicits.global
+
+@RestController
+class KafkaController @Inject()(kafkaService: KafkaService) {
+  @GetMapping(path = Array("/doesKafkaTopicExist"))
```
kevinwallimann: I think this endpoint does not follow best practices for REST :) I think we could get inspiration from Confluent Control Center, and name it like this: `/topics/{topic-name}/exists?servers=`

Obviously this is still not the textbook REST way, but we don't want to expose any other information, so I think, an exists endpoint is ok.

src/main/resources/application.properties
```diff
@@ -69,6 +69,7 @@ spring.mail.port=
 #Kafka sensor properties.
 kafkaSource.group.id.prefix=hyper_drive
 kafkaSource.poll.duration=500
+kafkaSource.connectionTimeoutMillis=10000
```
kevinwallimann, line 72: I think it's problematic to group this together with the configuration for the notification topic. The Kafka service and the kafka sensor share nothing in common except they both connect to some kafka server

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaService.scala
```diff
@@ -0,0 +1,67 @@
...
+
+trait KafkaService {
+  def doesKafkaTopicExist(kafkaTopicName: String, servers: Seq[String])(implicit ec: ExecutionContext): Future[Boolean]
+}
+
+@Service
+class KafkaServiceImpl(val kafkaConfig: KafkaConfig) extends KafkaService {
+  private val serviceLogger = LoggerFactory.getLogger(this.getClass)
+
+  override def doesKafkaTopicExist(kafkaTopicName: String, servers: Seq[String])(implicit ec: ExecutionContext): Future[Boolean] = {
```
kevinwallimann: I'd rename that. How about `existsTopic`?

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaService.scala
```diff
@@ -0,0 +1,67 @@
...
+          closeAdminClient(adminClient)
+          topics.exists(_._2.name() == kafkaTopicName)
+        case Failure(exception) => exception.getCause match {
+          case _: UnknownTopicOrPartitionException =>
+            closeAdminClient(adminClient)
+            false
+          case _ =>
+            closeAdminClient(adminClient)
+            serviceLogger.error(s"Unexpected error occurred checking kafka topic = $kafkaTopicName, servers = $servers", exception)
+            throw new ApiException(GenericError("Could not establish connection."))
```
kevinwallimann: How about "Unexpected error occurred while connecting to kafka brokers"?

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaService.scala
```diff
@@ -0,0 +1,67 @@
...
+
+@Service
+class KafkaServiceImpl(val kafkaConfig: KafkaConfig) extends KafkaService {
+  private val serviceLogger = LoggerFactory.getLogger(this.getClass)
+
+  override def doesKafkaTopicExist(kafkaTopicName: String, servers: Seq[String])(implicit ec: ExecutionContext): Future[Boolean] = {
+    def closeAdminClient(adminClient: AdminClient): Unit = if (adminClient != null) adminClient.close()
+
+    val properties = kafkaConfig.properties
+    properties.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, servers.mkString(","))
```
kevinwallimann: Don't you need more properties, especially when connecting securely through https? And if yes, I think you'd have to accept them through the GET request, since you can't assume which properties are set in the workflow.

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaService.scala
jozefbakus: Done

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaService.scala
jozefbakus: Done

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/KafkaController.scala
jozefbakus: Fixed

PR #537, opened 2021-11-22
Feature/535 show job template usage

### LINKED ISSUE #535: Show job template usage
UX - Show job template usage

jozefbakus:
Implements: #535 

Show job template usage

### DIFF
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/JobTemplateService.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/JobTemplateService.scala
@@ -17,7 +17,7 @@ package za.co.absa.hyperdrive.trigger.api.rest.services
 
 import org.springframework.stereotype.Service
 import za.co.absa.hyperdrive.trigger.models.search.{TableSearchRequest, TableSearchResponse}
-import za.co.absa.hyperdrive.trigger.models.{DagDefinitionJoined, JobTemplate, ResolvedJobDefinition}
+import za.co.absa.hyperdrive.trigger.models.{DagDefinitionJoined, JobTemplate, ResolvedJobDefinition, Workflow}
 import za.co.absa.hyperdrive.trigger.persistance.JobTemplateRepository
 
 import scala.concurrent.{ExecutionContext, Future}
@@ -35,6 +35,7 @@ trait JobTemplateService {
   def createJobTemplate(jobTemplate: JobTemplate)(implicit ec: ExecutionContext): Future[JobTemplate]
   def updateJobTemplate(jobTemplate: JobTemplate)(implicit ec: ExecutionContext): Future[JobTemplate]
   def deleteJobTemplate(id: Long)(implicit ec: ExecutionContext): Future[Boolean]
+  def getJobTemplateUsage(id: Long): Future[Seq[Workflow]]
 }
 
 @Service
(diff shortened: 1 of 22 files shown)


src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/JobTemplateService.scala
```diff
@@ -35,6 +35,7 @@ trait JobTemplateService {
   def createJobTemplate(jobTemplate: JobTemplate)(implicit ec: ExecutionContext): Future[JobTemplate]
   def updateJobTemplate(jobTemplate: JobTemplate)(implicit ec: ExecutionContext): Future[JobTemplate]
   def deleteJobTemplate(id: Long)(implicit ec: ExecutionContext): Future[Boolean]
+  def getJobTemplateUsage(id: Long): Future[Seq[Workflow]]
```
kevinwallimann: I would name this `getWorkflowsByJobTemplate`, similar to `getWorkflowsByProjectName`. You can then ask yourself, whether this belongs to the `WorkflowService` or the `JobTemplateService`, but for me it's ok here.

ui/src/app/components/admin/job-templates/job-templates-home/job-template-usage/job-template-usage.component.html
```diff
@@ -0,0 +1,38 @@
+<!--
```
kevinwallimann, line 1: This does look similar to the workflow grid. I feel a bit uneasy to "duplicate" this here and I'm wondering if it wouldn't be possible to reuse it? I mean, what are you going to do with that information? Wouldn't you want to deactivate, or export the listed workflows? Maybe we could add a workflows endpoint that takes the job template as a parameter, similar to the runs/{workflow-id} endpoint?

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/JobTemplateService.scala
jozefbakus: Renamed

ui/src/app/components/admin/job-templates/job-templates-home/job-template-usage/job-template-usage.component.html
jozefbakus: I am not sure about this. What this shows: usage of the job template - where is used and if it is used actively.
With the current implementation, it is better to have it "duplicated" because some methods on table uses workflow actions.
I would rather go with a new workflows endpoint.

kevinwallimann: Maybe let me rephrase this. In principle, you could also add a column on the workflow table to find the usage of the job templates (It wouldn't be quite so easy because the job templates are on the dag definition). That's why for me, the job templates usage, is just a list of workflows. And I believe over time, we will want to do the same things on this list of workflows as on the main workflow grid.

jozefbakus: I would leave it on the templates page. 
The workflows main page requires some refactoring. First, we need to implement backend filtering and sorting and replace the left side list with only project list and after that we can enhance table with new column

kevinwallimann (approved):

PR #545, opened 2021-12-07
Implement recurring sensor throttling

### LINKED ISSUE #540: Add throttling for recurring sensor
# Background

The recurring sensor was created to ensure a long-running job will immediately be restarted, should it unexpectedly fail due to intermittent issues that can be fixed with an application restart.

However, some jobs with the recurring sensor fail because of misconfiguration or they succeed but are short-lived jobs. In both cases, new job instances will be submitted very often, and more often than they should. In case they fail because of misconfiguration, the configuration should be fixed. In case of short-lived jobs, they should be converted to long-running jobs or scheduled using a cron expression with the Time Sensor.

To limit excessive job executions through the recurring sensor, a global maximum number of jobs per period should be implemented per recurring sensor. If the number of job invocations per period (e.g. 1 hour) is reached, no more job instances should be created for this sensor within this period. Manual invocations should still be possible. Activation / Deactivation should have no influence.

# Feature

- New application properties
  -  `recurringSensor.maxJobsPerPeriod` (accepting an Integer)
  -  `recurringSensor.period` (accepting a duration, see https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.external-config.typesafe-configuration-properties.conversion.durations)
- Add a query to `JobInstanceRepository` how many job instance repositories have been created in the past hour.
- Do not create `JobInstance` if that number is above the threshold (log the event)

kevinwallimann:
Closes #540 

- recurringSensor.maxJobsPerDuration (optional, default: 8)
- recurringSensor.duration (optional, default: 1h)

### DIFF
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/configuration/application/RecurringSensorConfig.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/configuration/application/RecurringSensorConfig.scala
@@ -1,0 +1,33 @@
+
+/*
+ * Copyright 2018 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.hyperdrive.trigger.configuration.application
+
+import org.springframework.boot.context.properties.bind.DefaultValue
+import org.springframework.boot.context.properties.{ConfigurationProperties, ConstructorBinding}
+import org.springframework.validation.annotation.Validated
+
+import java.time.Duration
+
+@ConfigurationProperties("recurring-sensor")
+@ConstructorBinding
+@Validated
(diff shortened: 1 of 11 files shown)


src/main/scala/za/co/absa/hyperdrive/trigger/configuration/application/RecurringSensorConfig.scala
```diff
@@ -0,0 +1,33 @@
...
+import org.springframework.boot.context.properties.{ConfigurationProperties, ConstructorBinding}
+import org.springframework.validation.annotation.Validated
+
+import java.time.Duration
+
+@ConfigurationProperties("recurring-sensor")
+@ConstructorBinding
+@Validated
+class RecurringSensorConfig (
+  @DefaultValue(Array("8"))
```
jozefbakus, line 29: How did you get this value ?

src/main/scala/za/co/absa/hyperdrive/trigger/configuration/application/RecurringSensorConfig.scala
```diff
@@ -0,0 +1,33 @@
...
+  val maxJobsPerDuration: Int,
+  @DefaultValue(Array("1h"))
+  val duration: Duration
```
jozefbakus, line 32: I would rather use Int value (minutes) instead of Duration

jozefbakus (approved):

src/main/scala/za/co/absa/hyperdrive/trigger/configuration/application/RecurringSensorConfig.scala
jozefbakus: empirically derived constant :D

Issue #563, opened 2021-12-13
Workflows main page refactoring - side bar 


jozefbakus:
Load All workflow/project names for a side bar in separate backend call

- [x] Backend: Implement get project and workflow names
- [x] Frontend: Workflows side bar refactoring

PR #578, opened 2021-12-21
Workflows main page refactoring - side bar

jozefbakus:
Implements: #563

### DIFF
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/persistance/WorkflowRepository.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/persistance/WorkflowRepository.scala
@@ -43,7 +43,8 @@ trait WorkflowRepository extends Repository {
   def updateWorkflow(workflow: WorkflowJoined, user: String)(implicit ec: ExecutionContext): Future[Unit]
   def switchWorkflowActiveState(id: Long, user: String)(implicit ec: ExecutionContext): Future[Unit]
   def updateWorkflowsIsActive(ids: Seq[Long], isActiveNewValue: Boolean, user: String)(implicit ec: ExecutionContext): Future[Unit]
-  def getProjects()(implicit ec: ExecutionContext): Future[Seq[String]]
+  def getProjects()(implicit ec: ExecutionContext): Future[Seq[Project]]
+  def getProjectNames()(implicit ec: ExecutionContext): Future[Seq[String]]
   def getProjectsInfo()(implicit ec: ExecutionContext): Future[Seq[ProjectInfo]]
   def existsProject(project: String)(implicit ec: ExecutionContext): Future[Boolean]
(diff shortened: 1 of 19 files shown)


src/main/scala/za/co/absa/hyperdrive/trigger/persistance/WorkflowRepository.scala
```diff
@@ -46,7 +46,8 @@ trait WorkflowRepository extends Repository {
   def updateWorkflow(workflow: WorkflowJoined, user: String)(implicit ec: ExecutionContext): Future[Unit]
   def switchWorkflowActiveState(id: Long, user: String)(implicit ec: ExecutionContext): Future[Unit]
   def updateWorkflowsIsActive(ids: Seq[Long], isActiveNewValue: Boolean, user: String)(implicit ec: ExecutionContext): Future[Unit]
-  def getProjects()(implicit ec: ExecutionContext): Future[Seq[String]]
+  def getProjects()(implicit ec: ExecutionContext): Future[Seq[Project]]
+  def getProjectNames()(implicit ec: ExecutionContext): Future[Seq[String]]
   def getProjectsInfo()(implicit ec: ExecutionContext): Future[Seq[ProjectInfo]]
```
kevinwallimann, line 51: Btw, is the `/workflows/projectsInfo` endpoint used anywhere?

ui/src/app/stores/workflows/workflows.effects.ts
```diff
@@ -51,38 +51,41 @@ export class WorkflowsEffects {
     private jobService: JobService,
     private store: Store<AppState>,
     private router: Router,
-    private toastrService: ToastrService,
+    public toastrService: ToastrService,
```
kevinwallimann: Why is this `public`?

kevinwallimann:
I tested the UI. Looks good

ui/src/app/stores/workflows/workflows.effects.ts
jozefbakus: Oh s***.
I had a problem with "should return initialize workflows failure if workflowService.getJobTemplates responds with an error."
I tried to spy toastrService directly.
The problem was with the order.
I tested toaster call before I executed underTest.workflowsInitialize

src/main/scala/za/co/absa/hyperdrive/trigger/persistance/WorkflowRepository.scala
jozefbakus: Nope, I think the method itself was but I removed it.

jozefbakus: I don't think so, we can remove it

kevinwallimann: Ok, let's do it in another PR. Doesn't really have to do anything with this task

kevinwallimann (approved):

PR #579, opened 2021-12-22
Spring boot upgrade

### LINKED ISSUE #539: Upgrade Spring Boot to v2.6.x, Spring Security Ldap to v5.6.x
With Spring Security v5.6.0, bug https://github.com/spring-projects/spring-security/pull/9740 (see also https://github.com/spring-projects/spring-security/issues/8846) has been fixed. 
Spring Boot Starter Security v2.6.0 updated to Spring Security v5.6.0, likewise Spring Security Ldap v5.6.0.

This fixes the issue in hyperdrive-trigger that the JSESSIONID cookie is not deleted when logging out. This is because tomcat sets the JSESSIONID cookie with path `/hyperdrive_trigger`, but to delete the cookie, spring sends an empty JSESSIONID cookie with the path `/hyperdrive_trigger/` with a trailing slash. Since the paths are not the same, the client will not overwrite the existing JSESSIONID cookie. As per https://www.rfc-editor.org/rfc/rfc6265#section-4.1.2, clients will only overwrite a cookie if their name, domain, and path are the same.

jozefbakus:
Implements #539 

spring.version: 2.5.2 -> 2.6.1
spring.security.ldap.version: 5.5.1 -> 5.6.1

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -78,11 +78,11 @@
         <play-ahc-ws-standalone.version>2.0.2</play-ahc-ws-standalone.version>
         <jackson-module.version>2.11.0</jackson-module.version>
         <alchim31>3.2.0</alchim31>
-        <spring.version>2.5.2</spring.version>
+        <spring.version>2.6.1</spring.version>
         <quartz.version>2.3.2</quartz.version>
         <log4j.version>2.17.0</log4j.version>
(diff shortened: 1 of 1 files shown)


kevinwallimann (approved):
Cookie deletion on logout works now.

Issue #583, opened 2022-01-10
Job template/workflow/notification history - Revert to previous version


jozefbakus:
Job template/workflow/notification history - Revert to the previous version - via update page

PR #585, opened 2022-01-13
Job template/Notification/Workflow revert

jozefbakus:
Implements #583 - Job template/Notification/Workflow revert

### DIFF
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/JobTemplateHistoryController.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/JobTemplateHistoryController.scala
@@ -32,6 +32,11 @@ class JobTemplateHistoryController @Inject()(jobTemplateHistoryService: JobTempl
     jobTemplateHistoryService.getHistoryForJobTemplate(jobTemplateId).toJava.toCompletableFuture
   }
 
+  @GetMapping(path = Array("/historyJobTemplate"))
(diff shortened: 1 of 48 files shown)


src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/JobTemplateHistoryController.scala
```diff
@@ -32,6 +32,11 @@ class JobTemplateHistoryController @Inject()(jobTemplateHistoryService: JobTempl
     jobTemplateHistoryService.getHistoryForJobTemplate(jobTemplateId).toJava.toCompletableFuture
   }
 
+  @GetMapping(path = Array("/historyJobTemplate"))
+  def getHistoryJobTemplate(@RequestParam jobTemplateHistoryId: Long): CompletableFuture[JobTemplate] = {
```
kevinwallimann: I would rename this. If `jobTemplateHistory` means "the history of the job template", then `historyJobTemplate` means "the job template of the history", which makes no sense. You could go for `historicJobTemplate`, but since we already have `jobTemplatesFromHistory`, I would choose `jobTemplateFromHistory`.

Same for the repository and service methods.

jozefbakus: Fixed

kevinwallimann (dismissed):

kevinwallimann (approved):

PR #591, opened 2022-01-17
Scala upgrade 2.12

### LINKED ISSUE #582: Upgrade Scala Version
Upgrade Scala Version.
If possible, go with the latest scala version

jozefbakus:
Upgrade to scala 2.12
Implements #582 
Changes:
- Removed `scalaVersion` from `scala-maven-plugin` (Recommended in docs)
- Test updated: Instead of using execution context from AsyncFlatSpec, overrided with global execution context

Tested locally with yarn deployment and shell execution + notifications and templates

### DIFF
--- a/src/test/scala/za/co/absa/hyperdrive/trigger/api/rest/WorkflowControllerTest.scala
+++ b/src/test/scala/za/co/absa/hyperdrive/trigger/api/rest/WorkflowControllerTest.scala
@@ -31,10 +31,11 @@ import za.co.absa.hyperdrive.trigger.configuration.application.TestGeneralConfig
 import za.co.absa.hyperdrive.trigger.models.errors.ApiException
 import za.co.absa.hyperdrive.trigger.models.{Project, Workflow, WorkflowImportExportWrapper}
 
-import scala.concurrent.ExecutionContext.Implicits.global
-import scala.concurrent.Future
+import scala.concurrent.{ExecutionContext, Future}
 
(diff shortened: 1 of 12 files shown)


src/test/scala/za/co/absa/hyperdrive/trigger/api/rest/WorkflowControllerTest.scala
```diff
@@ -31,10 +31,11 @@ import za.co.absa.hyperdrive.trigger.configuration.application.TestGeneralConfig
 import za.co.absa.hyperdrive.trigger.models.errors.ApiException
 import za.co.absa.hyperdrive.trigger.models.{Project, Workflow, WorkflowImportExportWrapper}
 
-import scala.concurrent.ExecutionContext.Implicits.global
-import scala.concurrent.Future
+import scala.concurrent.{ExecutionContext, Future}
 
 class WorkflowControllerTest extends AsyncFlatSpec with Matchers with MockitoSugar with BeforeAndAfter {
+  override implicit def executionContext: ExecutionContext = scala.concurrent.ExecutionContext.Implicits.global
```
kevinwallimann, line 37: Probably we need to revisit the usage of the `AsyncFlatSpec` here. If we override its execution context, I guess we could just as well use `FlatSpec`. However, this is not a problem of upgrading to 2.12 and was like this before, so it's ok for me in this PR

kevinwallimann (approved):

PR #600, opened 2022-01-24
#599: InProcessLauncher

### LINKED ISSUE #599: Replace SparkLauncher with InProcessLauncher
Replace SparkLauncher with InProcessLauncher

jozefbakus:
Implements #599: InProcessLauncher
Replaced dependency: spark-launcher with spark-yarn
Replaced InProcessLauncher with SparkLauncher

Removed properties: 
- `sparkYarnSink.hadoopConfDir`
- `sparkYarnSink.sparkHome`

Replaced with ENV VARIABLES:
- `HADOOP_CONF_DIR`
- `SPARK_HOME`


kevinwallimann:
One additional problem I see, is we have to execute the `<spark-home>/conf/spark-env.sh` script to load the environment variables. We might have to do this as part of the application startup

kevinwallimann (commented):
We will have to try this out on a test system. Therefore, I wouldn't want to merge this into develop like this, as we lose the `SparkLauncher`. So either we leave it as a long-living feature branch and update it periodically, or you keep the existing `SparkLauncher` and create a new implementation of `SparkClusterService`

Issue #603, opened 2022-01-31
Sort jobs in workflow by order


jozefbakus:
Sort jobs in workflow by order

PR #604, opened 2022-01-31
Fix - sort jobs in workflow

jozefbakus:
Implements #603 
Front end sorting + backend sorting


src/main/scala/za/co/absa/hyperdrive/trigger/persistance/WorkflowRepository.scala
```diff
@@ -151,7 +151,7 @@ class WorkflowRepositoryImpl @Inject()(
             sensor = s,
             dagDefinitionJoined = DagDefinitionJoined(
               workflowId = dd.workflowId,
-              jobDefinitions = wsddjd.map(_._4),
+              jobDefinitions = wsddjd.map(_._4).sortBy(_.order),
```
kevinwallimann, line 154: You can create a test case for this as well

kevinwallimann (approved):

PR #616, opened 2022-02-17
611: Fix - Expand side menu workflows/project when workflow is open

### LINKED ISSUE #611: Expand side menu workflows/project when workflow the from project is opened
Expand side menu workflows/project when workflow from the project is opened.

jozefbakus:
Implements #611

Expand side menu workflows/project when workflow from the project is opened.

### DIFF
--- a/ui/src/app/components/workflows/workflows.component.spec.ts
+++ b/ui/src/app/components/workflows/workflows.component.spec.ts
@@ -16,12 +16,17 @@
 import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
 import { WorkflowsComponent } from './workflows.component';
 import { provideMockStore } from '@ngrx/store/testing';
-import { ProjectModelFactory } from '../../models/project.model';
+import { ProjectModelFactory, WorkflowIdentityModelFactory } from '../../models/project.model';
 import { WorkflowModelFactory } from '../../models/workflow.model';
+import { RouterTestingModule } from '@angular/router/testing';
+import { Router, Routes } from '@angular/router';
 
 describe('WorkflowsComponent', () => {
   let fixture: ComponentFixture<WorkflowsComponent>;
   let underTest: WorkflowsComponent;
+  let router: Router;
+
+  const routes = [{ path: 'workflows/show/:id', component: {} }] as Routes;
 
   const initialAppState = {
     workflows: {
@@ -42,7 +47,9 @@ describe('WorkflowsComponent', () => {
       TestBed.configureTestingModule({
         providers: [provideMockStore({ initialState: initialAppState })],
         declarations: [WorkflowsComponent],
+        imports: [RouterTestingModule.withRoutes(routes)],
       }).compileComponents();
+      router = TestBed.inject(Router);
     }),
   );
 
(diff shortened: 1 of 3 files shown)


ui/src/app/components/workflows/workflows.component.ts
```diff
@@ -45,6 +47,27 @@ export class WorkflowsComponent implements AfterViewInit, OnDestroy {
...
   }
 
+  isWorkflowHighlighted(id: number): boolean {
+    return this.router.url.split('/').some((part) => part === id.toString());
+  }
+
+  isProjectClosed(project: string, workflows: WorkflowIdentityModel[]): boolean {
+    const isOpen =
+      this.openedProjects.has(project) || workflows.some((workflow: WorkflowIdentityModel) => this.isWorkflowHighlighted(workflow.id));
+    if (isOpen && !this.openedProjects.has(project)) {
```
kevinwallimann: Little bit redundant logic. If you expand this, it reads
`(A || B) && !A`, which is equivalent to `(A && !A) || (B && !A)`, which is equivalent to `B && !A`,
where `A = this.openedProjects.has(project)`, `B = workflows.some((workflow: WorkflowIdentityModel) => this.isWorkflowHighlighted(workflow.id))`.

So I would rewrite this if-condition like this:
`if(!this.openedProjects.has(project) && workflows.some((workflow: WorkflowIdentityModel) => this.isWorkflowHighlighted(workflow.id))) {`

ui/src/app/components/workflows/workflows.component.ts
```diff
@@ -45,6 +47,27 @@ export class WorkflowsComponent implements AfterViewInit, OnDestroy {
...
+      this.openedProjects.add(project);
+    }
+    return !isOpen;
```
kevinwallimann: Here, you could then just return `this.openedProjects.has(project)`

Imo, the logic is a bit easier to follow than the detour over `isOpen`

ui/src/app/components/workflows/workflows.component.spec.ts
```diff
@@ -65,4 +72,106 @@ describe('WorkflowsComponent', () => {
...
+        const idNonMatching = 5;
+        router.navigate(['workflows/show', 555]).then(() => {
+          expect(underTest.isWorkflowHighlighted(idNonMatching)).toBeFalse();
+        });
+      }),
+    );
+  });
+
+  it(
+    'openCloseProject() should toggle a project',
```
kevinwallimann: nitpick: `toggleProject` instead of `openCloseProject` might be a bit easier to understand

ui/src/app/components/workflows/workflows.component.ts
jozefbakus: Should be !this.openedProjects.has(project). Fixed.

ui/src/app/components/workflows/workflows.component.ts
jozefbakus: Nice, thanks. Fixed.

ui/src/app/components/workflows/workflows.component.spec.ts
jozefbakus: Yes, you are right. Fixed

kevinwallimann (approved):

PR #618, opened 2022-02-18
602: Workflows are shown to be unselected when moving across pages

### LINKED ISSUE #602: Workflows are shown to be unselected when moving across pages
1) If you select all, then switch to page 2, select all, and then switch back to page 1, the items are not selected
2) So you can select them again. That's how that counter can end up with more items than total items.

image

Discussed solutions: 
- [x]  Try to upgrade clarity
- [x]  If not possible to fix it, disable selection on multiple pages.

jozefbakus:
Implements #602

As discussed: If not possible to fix it, disable selection on multiple pages.


kevinwallimann (approved):
When you have a selection and then choose to display more rows, the selection vanishes. That's unfortunate, however, the current version does not behave correctly in that case either, so it's ok for me.

PR #663, opened 2022-04-21
#659: Reduce logging in hyperdrive

jozefbakus:
Implements #659

### DIFF
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/Executors.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/Executors.scala
@@ -91,7 +91,7 @@ class Executors @Inject()(dagInstanceRepository: DagInstanceRepository, jobInsta
           }
         }
         fut.onComplete {
-          case Success(_) => logger.debug(s"Executing job. Job instance id = ${jobInstance}")
+          case Success(_) => logger.debug(s"Executing job. Job instance = ${jobInstance}")
           case Failure(exception) => {
             logger.error(s"Executing job failed. Job instance id = ${jobInstance}.", exception)
           }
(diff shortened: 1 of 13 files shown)


src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/spark/SparkEmrClusterServiceImpl.scala
```diff
@@ -62,8 +62,8 @@ class SparkEmrClusterServiceImpl @Inject()(
 
       val response = emr.addJobFlowSteps(jobFlowStepsRequest)
       val stepId = response.getStepIds.asScala.headOption
-      logger.info(s"Added jobFlowStepsRequest ${jobFlowStepsRequest} for executorId ${id} and stepId $stepId}")
-      logger.info(response.toString)
+      logger.debug(s"Added jobFlowStepsRequest ${jobFlowStepsRequest} for executorId ${id} and stepId $stepId}")
+      logger.debug(response.toString)
```
kevinwallimann: I'd like to keep this as `.info` because it's only during submission, so should not cause excessive logging

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/Executors.scala
```diff
@@ -101,6 +101,7 @@ class Executors @Inject()(dagInstanceRepository: DagInstanceRepository, jobInsta
   }
 
   private def updateJob(jobInstance: JobInstance): Future[Unit] = {
+    logger.info(s"Job updated. ID = ${jobInstance.id} STATUS = ${jobInstance.jobStatus} EXECUTOR_ID = ${jobInstance.executorJobId}")
```
kevinwallimann, line 104: Slightly off-topic, but maybe we could add a second parameter `oldJobInstance` and only update if the entity has changed?

kevinwallimann (commented):
How about changing the log level to info in `src/main/resources/logback.xml`?

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/spark/SparkEmrClusterServiceImpl.scala
jozefbakus: Done

jozefbakus:
> How about changing the log level to info in `src/main/resources/logback.xml`?

Done

kevinwallimann (approved):

PR #684, opened 2022-06-10
#672 - Added license checker

jozefbakus:
Implements #672 - Added license checker


pom.xml
```diff
@@ -636,6 +620,53 @@
                     <commitIdGenerationMode>full</commitIdGenerationMode>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.rat</groupId>
+                <artifactId>apache-rat-plugin</artifactId>
+                <version>0.13</version>
```
kevinwallimann: You don't need the version, since it's taken from the root pom

jozefbakus: Removed

kevinwallimann (dismissed):

PR #697, opened 2022-06-23
Feature/687 hdfs service

### LINKED ISSUE #687: Add service to read checkpoint folder
**Background**
The goal is to prevent Hyperdrive job instance creations when there are no messages available on a kafka topic (or only a few). To that end, the offsets from the checkpoint folder should be compared with the latest offsets from the kafka topic.

This issue is about getting the offsets from the checkpoint folder.

All configurations (including keytab location) needed to access the checkpoint folder should be available from the resolved job definition.

**Task**
A service method should be written, that takes a `ResolvedJobDefinition` and returns the latest offset from the checkpoint. The method should also return whether or not the offset has been committed.

**Related**
#688, #690

`kafka.consumers.cache.size` Optional, default 50

### LINKED ISSUE #688: Add service to read latest available offset from kafka topic
**Background**
The goal is to prevent Hyperdrive job instance creations when there are no messages available on a kafka topic (or only a few). To that end, the offsets from the checkpoint folder should be compared with the latest offsets from the kafka topic.

This issue is about getting the offsets from the kafka topic.

All configurations (including keytab location, broker urls, certificate locations) needed to access the kafka topic should be available from the resolved job definition.

**Task**
A service method should be written, that takes a `ResolvedJobDefinition` and returns the latest offset from the kafka topic.

**Related**
#687, #690

### LINKED ISSUE #690: Add service to decide if a DagInstance creation can be skipped
**Background**
The goal is to prevent Hyperdrive job instance creations when there are no messages available on a kafka topic (or only a few). To that end, the offsets from the checkpoint folder should be compared with the latest offsets from the kafka topic.

This issue is about deciding whether a dag instance should be created or not based on the latest checkpoint offset and kafka offset.

**Task**
In the `DagInstanceService` a new dag instance is created. The creation of this dag instance can be skipped, if the latest kafka offset is equal to the latest checkpoint offset, if the latest checkpoint offset has been committed. If the latest checkpoint offset has not been committed, a dag instance has to be created.

**Related**
#687, #688

kevinwallimann:
Closes #688 
Closes #687 
Closes #690 

`kafka.consumers.cache.size` Optional, default 50


src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaService.scala
```diff
@@ -0,0 +1,59 @@
...
+import javax.inject.Inject
+import scala.collection.mutable
+
+trait KafkaService {
+  def getEndOffsets(topic: String, consumerProperties: Properties): Map[Int, Long]
+}
+
+@Service
+class KafkaServiceImpl @Inject() (kafkaConfig: KafkaConfig, generalConfig: GeneralConfig) extends KafkaService {
+  private val logger = LoggerFactory.getLogger(this.getClass)
```
jozefbakus, line 38: Not used

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaService.scala
```diff
@@ -0,0 +1,59 @@
...
+  private val kafkaConsumersCache: mutable.Map[Properties, KafkaConsumer[String, String]] = mutable.Map()
```
jozefbakus: What about concurrent map, this map will be accessed concurrently

kevinwallimann: Yes, absolutely. Not sure why I forgot about that

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HdfsService.scala
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  /**
+   *  @return an Option of a String, Boolean pair. The string contains the path to the latest offset file, while the
+   *         boolean is true if the offset is committed (i.e. a corresponding commit file exists), and false otherwise.
+   *         None is returned if the offset file does not exist. If the offset file does not exist, the corresponding
+   *         commit file is assumed to also not exist.
+   */
+  override def getLatestOffsetFilePath(params: HdfsParameters): Option[(String, Boolean)] = {
+    userGroupInformationWrapper.loginUserFromKeytab(params.principal, params.keytab)
```
jozefbakus: Why is this needed ? 
There is a method parseFileAndClose which does not do it but reads file

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HdfsService.scala
```diff
@@ -0,0 +1,167 @@
...
+    val offsetBatchIdOpt = getLatestOffsetBatchId(params.checkpointLocation)
```
jozefbakus: Is this a good logic ? If I do understand correctly. 
You get latest offset and then you try to find commit for it. 
Why don't you find latest commit and based on that get an offset file ?

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveOffsetComparisonService.scala
```diff
@@ -0,0 +1,196 @@
...
+      if (kafkaParameters.isEmpty) {
+        logger.warn(
+          s"Could not find required kafka parameters in job definition ${jobDefinition.name} with args ${args}"
+        )
+      }
+      kafkaParameters
+    }
+  }
+
+  private def isHyperdriveJob(jobDefinition: ResolvedJobDefinition) =
```
jozefbakus: I don't get this method. 
Should not it be obDefinition.jobParameters.jobType == JobTypes.Hyperdrive &&
      jobDefinition.jobParameters.isInstanceOf[SparkInstanceParameters]

jozefbakus (commented):
As we discussed:
1. Get kafka offset first (Should be faster than HDFS) or do it in parallel 
2. If topic does not exist or offset is 0, do not job.
3. Kafka call errors are not handled

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HdfsService.scala
kevinwallimann: It's certainly needed, since you access Hdfs. I forgot to add it in parseFileAndClose

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HdfsService.scala
kevinwallimann: If the latest offset file doesn't exist, then there's no point in searching for the latest commit file, so I don't think finding the commit file first is always better. I don't know what will be more frequent

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveOffsetComparisonService.scala
kevinwallimann: It should be `isNotHyperdriveJob` 🤦 Will change the logic as you suggested and negate when I call it

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaService.scala
kevinwallimann: Done

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaService.scala
kevinwallimann: Done

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveOffsetComparisonService.scala
kevinwallimann: Done

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HdfsService.scala
kevinwallimann: Done

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaService.scala
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  override def getEndOffsets(topic: String, consumerProperties: Properties): Map[Int, Long] = {
+    getOffsets(topic, consumerProperties, EndOffsets)
+  }
+
+  def createKafkaConsumer(properties: Properties) = new KafkaConsumer[String, String](properties)
+
+  private def getOffsets(topic: String, properties: Properties, offsetFn: OffsetFunction): Map[Int, Long] = {
+    val groupId = s"${kafkaConfig.groupIdPrefix}-${generalConfig.appUniqueId}-kafkaService"
```
jozefbakus: I think it is not necessary to have kafka and general config here, just any uuid should be fine.

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaService.scala
```diff
@@ -0,0 +1,80 @@
...
+
+trait KafkaService {
+  def getBeginningOffsets(topic: String, consumerProperties: Properties): Map[Int, Long]
+  def getEndOffsets(topic: String, consumerProperties: Properties): Map[Int, Long]
+}
+
+@Service
+class KafkaServiceImpl @Inject() (kafkaConfig: KafkaConfig, generalConfig: GeneralConfig) extends KafkaService {
+
+  private val kafkaConsumersCache = new ConcurrentHashMap[Properties, KafkaConsumer[String, String]]()
```
jozefbakus: Using just hash map may lead to out of mem issue.
If you update a workflow and change kafka sensor to time, consumer will stay in memory (also if you just change kafka sensor conf). 
If all instances go down except one, that instance will keep in mem consumers for all kafka sensors.

I think at least some cache with time expiration is needed or just create/destroy consumer every time

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaService.scala
```diff
@@ -0,0 +1,80 @@
...
+import javax.inject.Inject
+import scala.collection.JavaConverters._
+
+trait KafkaService {
+  def getBeginningOffsets(topic: String, consumerProperties: Properties): Map[Int, Long]
+  def getEndOffsets(topic: String, consumerProperties: Properties): Map[Int, Long]
+}
+
+@Service
+class KafkaServiceImpl @Inject() (kafkaConfig: KafkaConfig, generalConfig: GeneralConfig) extends KafkaService {
```
jozefbakus, line 35: We take all security configurations from KafkaConfig, but I could not find place where is that used

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HdfsService.scala
```diff
@@ -0,0 +1,171 @@
...
+import org.apache.hadoop.fs.{FileSystem, Path, PathFilter}
+import org.apache.spark.deploy.SparkHadoopUtil
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Service
+
+import java.nio.charset.StandardCharsets.UTF_8
+import javax.inject.Inject
+import scala.io.Source
+
+trait HdfsService {
```
jozefbakus, line 30: This is not just HdfsService, it can also parse file and  it understands spark structured streaming.

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/UserGroupInformationWrapper.scala
```diff
@@ -0,0 +1,30 @@
...
+import org.springframework.stereotype.Service
+
+trait UserGroupInformationWrapper {
+  def loginUserFromKeytab(principal: String, keytab: String): Unit
+}
+
+@Service
+class UserGroupInformationWrapperImpl extends UserGroupInformationWrapper {
+  override def loginUserFromKeytab(principal: String, keytab: String): Unit =
+    UserGroupInformation.loginUserFromKeytab(principal, keytab)
```
jozefbakus, line 29: Every time it tries to read, it logins and refreshes ticket. 
That is not necessary, you can check if ticket is valid if yes just continue

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HdfsService.scala
jozefbakus: loginUserFromKeytab should not be a public interface and it should be used only in this class.
Also methods parseKafkaOffsetStream and getLatestOffsetFilePath should be somewhere else, if you want to have a hdfs service, it should contain only read file, list dirs, list files and so on

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveOffsetComparisonService.scala
```diff
@@ -0,0 +1,252 @@
...
+import za.co.absa.hyperdrive.trigger.models.enums.JobTypes
+import za.co.absa.hyperdrive.trigger.models.{JobInstanceParameters, SparkInstanceParameters}
+
+import java.util.Properties
+import javax.inject.Inject
+import scala.concurrent.{ExecutionContext, Future}
+
+trait HyperdriveOffsetComparisonService {
+  def getResolvedAppArguments(jobParameters: JobInstanceParameters): Option[Map[String, String]]
+  def getHdfsParameters(resolvedAppArguments: Map[String, String]): Option[HdfsParameters]
```
jozefbakus: What is your plan for getHdfsParameters,  getKafkaParameters ?
Maybe util class could be better place for them

jozefbakus: It will keep this class a bit cleaner

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaService.scala
```diff
@@ -0,0 +1,80 @@
...
+
+  override def getBeginningOffsets(topic: String, consumerProperties: Properties): Map[Int, Long] = {
+    getOffsets(topic, consumerProperties, BeginningOffsets)
+  }
+
+  override def getEndOffsets(topic: String, consumerProperties: Properties): Map[Int, Long] = {
+    getOffsets(topic, consumerProperties, EndOffsets)
+  }
+
+  def createKafkaConsumer(properties: Properties) = new KafkaConsumer[String, String](properties)
```
jozefbakus: This method is not needed you can just create new consumer

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HdfsService.scala
jozefbakus: I thinking more like SparkMetadataService, with one public interface: give me offsets

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaService.scala
kevinwallimann: I need it for testing

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HdfsService.scala
kevinwallimann: Yeah need to rethink this. Tbh, the other methods are only public for simplified testing

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaService.scala
```diff
@@ -0,0 +1,79 @@
...
+  override def getBeginningOffsets(topic: String, consumerProperties: Properties): Map[Int, Long] = {
+    getOffsets(topic, consumerProperties, BeginningOffsets)
+  }
+
+  override def getEndOffsets(topic: String, consumerProperties: Properties): Map[Int, Long] = {
+    getOffsets(topic, consumerProperties, EndOffsets)
+  }
+
+  def createKafkaConsumer(propertiesThreadId: (Properties, Long)) =
+    new KafkaConsumer[String, String](propertiesThreadId._1)
```
jozefbakus, line 57: Could you please add debug log with propertiesHash or better would be kafka topic name and thread id

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/CheckpointService.scala
```diff
@@ -42,8 +46,7 @@ class HdfsParameters(
 )
 
 @Service
-class CheckpointServiceImpl @Inject() (userGroupInformationWrapper: UserGroupInformationWrapper)
-    extends CheckpointService {
+class CheckpointServiceImpl @Inject() (hdfsService: HdfsService) extends CheckpointService {
```
jozefbakus, line 45: private lazy val fs = FileSystem.get(conf) and 
private lazy val conf = SparkHadoopUtil.get.conf and
imports
not used

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/CheckpointService.scala
```diff
@@ -104,23 +109,29 @@ class CheckpointServiceImpl @Inject() (userGroupInformationWrapper: UserGroupInf
...
-        case e: Exception =>
-          // re-throw the exception with the log file path added
-          throw new Exception(s"Failed to parse file $path", e)
-      } finally {
-        IOUtils.closeQuietly(input)
+    hdfsService.exists(path).flatMap { exists =>
+      if (exists) {
+        hdfsService.open(path).map { input =>
+          try {
+            val lines = Source.fromInputStream(input, UTF_8.name()).getLines()
```
jozefbakus: I would leave open and read on hdfs service -> hdfsService.parseFileAndClose

src/main/resources/application.properties
```diff
@@ -70,6 +70,9 @@ notification.sender.address=
 spring.mail.host=
 spring.mail.port=
 
+# Kafka service properties.
+kafka.consumers.cache.size=50
```
jozefbakus, line 74: Could you please add this into PR description, easier to scrap it during release

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/CheckpointService.scala
kevinwallimann: I knew I forgot something 🤦

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/CheckpointService.scala
kevinwallimann: Ok, yes makes sense

jozefbakus (dismissed):
👍🏻

jozefbakus (approved):

PR #702, opened 2022-06-28
#698 - Add support for extra yarn tags

### LINKED ISSUE #698: Add support for extra yarn tags
# Background
Currently, `spark.yarn.tags` is set in the backend with a generated UUID.
Therefore, if `spark.yarn.tags` is set in the workflow as an extra spark configuration, it would be overwritten by the generated UUID.

# Feature
The user-defined (or template-defined) `spark.yarn.tags` should be merged with the generated UUID in the backend.

`spark.yarn.tags` is a comma-separated string, e.g. `spark.yarn.tags=tag1,tag2,tag3`

jozefbakus:
Implements: #698


src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/spark/SparkClusterService.scala
```diff
@@ -32,10 +32,20 @@ trait SparkClusterService {
...
+    val extraJavaOptionsMerge = SparkExtraJavaOptions.KeysToMerge.map { key =>
       val globalValue = globalConfig.getOrElse(key, "")
       val jobValue = jobConfig.getOrElse(key, "")
-      key -> s"$globalValue$MergedValuesSeparator$jobValue".trim
-    }.toMap
+      key -> s"$globalValue${SparkExtraJavaOptions.MergedValuesSeparator}$jobValue".trim
+    }
+    val tagsOptions = SparkTags.KeysToMerge.map { key =>
+      val globalValue = globalConfig.get(key)
+      val jobValue = jobConfig.get(key)
```
kevinwallimann, line 43: Add some test code, or try to reuse the logic in `JobTemplateResolutionService`. For example, you could use `.getOrElse(Array.empty)` here.

jozefbakus: I added tests.
I cannot use .getOrElse(Array.empty) because, I need a string.
This gives me Option which I can filter out if is None. 
I could use .getOrElse("") but this has to be filtered anyway because if I split "" it gives me an array with one empty element, so I would rather have Optional rather then String and EmptyString

kevinwallimann (dismissed):

PR #707, opened 2022-07-26
Feature/700 skip dag instance 2

### LINKED ISSUE #700: Skip dag instance creation if no new message is available in kafka
# Feature
- Add logic in `Executors` (see around lines 100+)
- For Hyperdrive jobs, determine if there are messages to be consumed, using the service created in #690 HyperdriveOffsetComparisonService.isNewJobInstanceRequired
- Create a HyperdriveExecutor, copy execute method from SparkExecutor, job status must be updated always (cannot be skipped)
- If no message is available, finish stop with status `No Data`
- If message is available, use the `SparkExecutor`

kevinwallimann:
Closes #700 

Sets Dag instance status to Skipped if all job instances have status NoData (or if there are zero job instances).

**Added a property for testing**

`executors.enable.hyperdrive.executor` (optional, default true). If set to false, HyperdriveExecutor will not be used and instead, the SparkExecutor will be used for all workflows.

On top of enabling the HyperdriveExecutor globally, jobs also need to opt-in by specifying the following app argument:
`useHyperdriveExecutor=true`

Both items are only intended during a testing period, after which the logic for both things should be removed.

**Symbolic links on tomcat**
See section "Symbolic links on user-defined files" in the README for when and why to add symbolic links in the tomcat home folder to keytab and keystores.

### DIFF
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/CheckpointService.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/CheckpointService.scala
@@ -15,10 +15,10 @@
 
 package za.co.absa.hyperdrive.trigger.api.rest.services
 
-import com.fasterxml.jackson.databind.ObjectMapper
-import com.fasterxml.jackson.module.scala.DefaultScalaModule
 import org.apache.hadoop.fs.{Path, PathFilter}
 import org.apache.hadoop.security.UserGroupInformation
+import org.json4s.jackson.Serialization
+import org.json4s.{Formats, NoTypeHints}
 import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Service
 import za.co.absa.hyperdrive.trigger.api.rest.utils.ScalaUtil.swap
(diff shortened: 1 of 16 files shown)


src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/CheckpointService.scala
```diff
@@ -101,11 +101,12 @@ class CheckpointServiceImpl @Inject() (hdfsService: HdfsService) extends Checkpo
    *  and org.apache.spark.sql.kafka010.JsonUtils
    *  for details on the assumed format
    */
+
```
jozefbakus: Unnecessary :D

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaService.scala
```diff
@@ -58,7 +59,7 @@ class KafkaServiceImpl @Inject() (generalConfig: GeneralConfig) extends KafkaSer
   }
 
   private def getOffsets(topic: String, properties: Properties, offsetFn: OffsetFunction): Map[Int, Long] = {
-    val groupId = s"hyperdrive-trigger-kafkaService-${randomUUID().toString}"
+    val groupId = s"hyperdrive-trigger-kafkaService-$consumerUuid"
```
jozefbakus, line 62: I haven't noticed this 👍

jozefbakus (dismissed):

jozefbakus (commented):
Could you please add/document symlink

jozefbakus (approved):

PR #708, opened 2022-08-03
#703 - Recognize dead scheduler instances without relying on synchronized clocks

### LINKED ISSUE #703: Recognize dead scheduler instances without relying on synchronized clocks
DRAFT IN PROGRESS

# Background

Currently, schedulers are presumed dead / not responding when the latest timestamp.

Idea: Use current_timestamp of database instead of sending the local timestamps

jozefbakus:
Closes #703

Recognize dead scheduler instances without relying on synchronized clocks

### DIFF
--- a/src/test/scala/za/co/absa/hyperdrive/trigger/persistance/SchedulerInstanceRepositoryTest.scala
+++ b/src/test/scala/za/co/absa/hyperdrive/trigger/persistance/SchedulerInstanceRepositoryTest.scala
@@ -97,7 +97,7 @@ class SchedulerInstanceRepositoryTest
     val expectedBefore = LocalDateTime.now()
     val result = await(schedulerInstanceRepository.getCurrentDateTime())
     val expectedAfter = LocalDateTime.now()
-    result isAfter expectedBefore
-    result isBefore expectedAfter
+    (result isAfter expectedBefore) || (result equals expectedBefore) shouldBe true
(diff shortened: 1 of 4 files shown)


src/test/scala/za/co/absa/hyperdrive/trigger/persistance/SchedulerInstanceRepositoryTest.scala
```diff
@@ -92,4 +92,14 @@ class SchedulerInstanceRepositoryTest
     val result = await(schedulerInstanceRepository.getAllInstances())
     result should contain theSameElementsAs TestData.schedulerInstances
   }
+
+  "getCurrentDateTime" should "return database current date time" in {
+    val result = await(schedulerInstanceRepository.getCurrentDateTime())
+    val expected = LocalDateTime.now()
```
kevinwallimann: Hmm, you could get unlucky with that. What if `schedulerInstanceRepository.getCurrentDateTime()` is executed at e.g. 2022-12-31 23:59:59.999

What you could do is take the `LocalDateTime.now()` before invoking `schedulerInstanceRepository.getCurrentDateTime()` and after it and assert that the result is in between.

jozefbakus: Fixed

kevinwallimann (approved):

Issue #736, opened 2022-11-11
Kafka sensor - always catch up missed messages


jozefbakus:
Kafka sensor - always catch up missed messages (At least once for kafka sensors)

Remove isFirstIteration logic from scheduler 
Add property always catch up on missed messages

PR #741, opened 2022-11-23
Always catch up on missed messages 

jozefbakus:
Implements: #736 
Added new property: 
`kafkaSource.always.catchup` with default `true`

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ kafka.consumers.cache.size=50
 #Kafka sensor properties. Not all are required. Adjust according to your use case.
 kafkaSource.group.id.prefix=hyper_drive_${appUniqueId}
 kafkaSource.poll.duration=500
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -137,6 +137,7 @@ kafka.consumers.cache.size=50
 #Kafka sensor properties. Not all are required. Adjust according to your use case.
 kafkaSource.group.id.prefix=hyper_drive_${appUniqueId}
 kafkaSource.poll.duration=500
+kafkaSource.always.seek.to.end=false
```
kevinwallimann: I like `kafkaSource.always.catchup` better. I see why you named it `always.seek.to.end`, but it seems not straight-forward to understand

jozefbakus: Done

kevinwallimann (approved):

Issue #758, opened 2022-12-16
Fix flaky NotificationRuleValidationServiceTest


kevinwallimann:
## Describe the bug
The NotificationRuleValidationServiceTest sometimes fails. Apparently, this is because of two threads that attempt to run the suite at the same time. See https://www.scalatest.org/user_guide/async_testing

## To Reproduce
Does not always fail. See e.g. https://github.com/AbsaOSS/hyperdrive-trigger/actions/runs/3711658533/attempts/1

## Expected behavior
The test should always pass.

## Stacktraces / Screenshots
```
NotificationRuleValidationServiceTest:
validate
za.co.absa.hyperdrive.trigger.api.rest.services.NotificationRuleValidationServiceTest *** ABORTED ***
  java.util.ConcurrentModificationException: Two threads have apparently attempted to run a suite at the same time. This has resulted in both threads attempting to concurrently change the current Informer. Suite class name: za.co.absa.hyperdrive.trigger.api.rest.services.NotificationRuleValidationServiceTest
  at org.scalatest.AsyncSuperEngine.$anonfun$runImpl$5(AsyncEngine.scala:626)
  at org.scalatest.AsyncSuperEngine.$anonfun$runImpl$5$adapted(AsyncEngine.scala:623)
  at org.scalatest.CompositeStatus.whenCompleted(Status.scala:886)
  at org.scalatest.AsyncSuperEngine.runImpl(AsyncEngine.scala:623)
  at org.scalatest.AsyncFlatSpecLike.run(AsyncFlatSpecLike.scala:1812)
  at org.scalatest.AsyncFlatSpecLike.run$(AsyncFlatSpecLike.scala:1810)
  at za.co.absa.hyperdrive.trigger.api.rest.services.NotificationRuleValidationServiceTest.org$scalatest$BeforeAndAfter$$super$run(NotificationRuleValidationServiceTest.scala:30)
  at org.scalatest.BeforeAndAfter.run(BeforeAndAfter.scala:258)
  at org.scalatest.BeforeAndAfter.run$(BeforeAndAfter.scala:256)
  at za.co.absa.hyperdrive.trigger.api.rest.services.NotificationRuleValidationServiceTest.run(NotificationRuleValidationServiceTest.scala:30)
  ...
```

PR #765, opened 2023-01-04
Fix/758 fix flaky notification rule validation service test

jozefbakus:
Implements: #758

The error indicates two tests in the exact same suite instance are trying to be run concurrently. The model is that only one runs at a time.
So I did mix in ParallelTestExecution trait to make tests parallel


src/test/scala/za/co/absa/hyperdrive/trigger/api/rest/services/NotificationRuleValidationServiceTest.scala
```diff
@@ -18,7 +18,7 @@ package za.co.absa.hyperdrive.trigger.api.rest.services
 import org.mockito.ArgumentMatchers.{eq => eqTo, _}
 import org.mockito.Mockito._
 import org.scalatest.mockito.MockitoSugar
-import org.scalatest.{AsyncFlatSpec, BeforeAndAfter, Matchers}
+import org.scalatest.{AsyncFlatSpec, BeforeAndAfter, Matchers, ParallelTestExecution}
```
kevinwallimann: I don't think this is the way to go. We have mocks that are used by all test cases, such as `workflowRepository`. If the test cases are allowed to run in parallel, that introduces race conditions.

I think we should just stick to the documentation and wrap the assertions into future transformers (e.g. `map`) instead of using `Await`. Then we can also use the default single-thread execution context and don't need to override it with the global execution context, which is probably causing the problems.

Example:
```
class FutureTest extends AsyncFlatSpec {
  "A Future" should "return the expected result" in {
    val f = Future {
      Thread.sleep(100)
      "Hello"
    }
    f.map { result =>
      assert(result == "Hello")
    }
  }
}
```

jozefbakus: I forgot to move it to every test.
But I did it as documentation says.

kevinwallimann (approved):

Issue #752, opened 2022-12-08
Fix compiler warnings


kevinwallimann:
Fix compiler warnings

Feel free to create subissues for complicated fixes.

PR #777, opened 2023-01-27
Fix/752 fixed compiler warnings UI

jozefbakus:
Implements #752


ui/src/app/stores/application/application.effects.ts
```diff
@@ -25,27 +25,28 @@ import { AppInfoModel } from '../../models/appInfo.model';
 export class ApplicationEffects {
   constructor(private actions: Actions, private appInfoService: AppInfoService) {}
 
-  @Effect({ dispatch: true })
```
kevinwallimann, line 28: The dispatch true is not necessary anymore?

kevinwallimann (approved):

ui/src/app/stores/application/application.effects.ts
jozefbakus: No, only if it does not dispatch

Issue #789, opened 2023-02-10
Show user already ingested offset


jozefbakus:
## Background
For all hyperdrive type jobs, user should be able to see on what offsets ingestion already is. 

## Feature
For all hyperdrive type jobs, user should be able to see on what offsets ingestion already is.

PR #791, opened 2023-02-23
Feature/789 show messages to ingest on hyperdrive jobs

jozefbakus:
Implements: #789


src/main/scala/za/co/absa/hyperdrive/trigger/models/IngestionStatus.scala
```diff
@@ -0,0 +1,24 @@
...
+
+package za.co.absa.hyperdrive.trigger.models
+
+case class IngestionStatus(
+  jobName: String,
+  jobType: String,
+  topic: Option[Topic]
+)
+
+case class Topic(topic: String, messagesToIngest: Map[Int, Long])
```
kevinwallimann: Please rename the case class to `TopicStatus` or something. I was quite confused when reading the code.

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/CheckpointService.scala
```diff
@@ -98,6 +100,18 @@ class CheckpointServiceImpl @Inject() (@Lazy hdfsService: HdfsService) extends C
...
+  override def getLatestCommittedOffset(
+    params: HdfsParameters
+  )(implicit ugi: UserGroupInformation): Try[Option[Map[Int, Long]]] = {
+    getLatestCommitBatchId(params.checkpointLocation).map {
+      case Some(latestCommit) =>
+        val pathToLatestCommit = new Path(s"${params.checkpointLocation}/$offsetsDirName/$latestCommit")
+        getOffsetsFromFile(pathToLatestCommit.toString)
+          .map(_.map(topicPartitionOffsets => topicPartitionOffsets.head._2))
+      case None => Try(Option.empty[Map[Int, Long]])
+    }.flatten
```
kevinwallimann: Did you use map + flatten vs flatMap on purpose?

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/CheckpointService.scala
```diff
@@ -98,6 +100,18 @@ class CheckpointServiceImpl @Inject() (@Lazy hdfsService: HdfsService) extends C
...
   }
 
+  override def getLatestCommittedOffset(
+    params: HdfsParameters
+  )(implicit ugi: UserGroupInformation): Try[Option[Map[Int, Long]]] = {
+    getLatestCommitBatchId(params.checkpointLocation).map {
+      case Some(latestCommit) =>
+        val pathToLatestCommit = new Path(s"${params.checkpointLocation}/$offsetsDirName/$latestCommit")
+        getOffsetsFromFile(pathToLatestCommit.toString)
+          .map(_.map(topicPartitionOffsets => topicPartitionOffsets.head._2))
```
kevinwallimann: I understand now your question about supporting multiple topics. Probably I would have implemented it generically here in the checkpoint service, and then extracted the partition offsets for the first topic only in the hyperdriveOffsetService. There is no right and wrong, however now there is a method in this class which returns the partition offset for all topics (`getOffsetsFromFile`) and one that only returns it for the first topic.

It seems that this method belongs to `HyperdriveOffsetService`, it seems similar to `getCheckpointOffsets`

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveOffsetService.scala
```diff
@@ -52,6 +57,58 @@ class HyperdriveOffsetComparisonServiceImpl @Inject() (sparkConfig: SparkConfig,
   private val ListDelimiter = ','
   private val defaultDeserializer = "org.apache.kafka.common.serialization.StringDeserializer"
 
+  /**
+   *  @param jobParameters Parameters for the job instance. Should contain at least
+   *                      - reader.kafka.topic
+   *                      - reader.kafka.brokers
+   *                      - writer.common.checkpoint.location
+   *  @param ec            ExecutionContext
+   *  @return - number not ingested messages.
```
kevinwallimann: number of not ingested messages for each topic and partition

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveService.scala
```diff
@@ -0,0 +1,83 @@
...
+import za.co.absa.hyperdrive.trigger.models.enums.JobTypes
+import za.co.absa.hyperdrive.trigger.persistance.WorkflowRepository
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success}
+
+trait HyperdriveService {
+  val workflowRepository: WorkflowRepository
+  val jobTemplateService: JobTemplateService
+  val hyperdriveOffsetService: HyperdriveOffsetService
```
kevinwallimann: Why are these services public vals?

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/CheckpointService.scala
```diff
@@ -98,6 +100,18 @@ class CheckpointServiceImpl @Inject() (@Lazy hdfsService: HdfsService) extends C
...
+      case None => Try(Option.empty[Map[Int, Long]])
```
kevinwallimann: I would log a warning here

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveOffsetService.scala
```diff
@@ -52,6 +57,58 @@ class HyperdriveOffsetComparisonServiceImpl @Inject() (sparkConfig: SparkConfig,
...
+
+    Future(
+      for {
+        kafkaParameters <- kafkaParametersOpt
+        hdfsParameters <- hdfsParametersOpt
+      } yield {
+        val kafkaOffsets = kafkaService.getOffsets(kafkaParameters._1, kafkaParameters._2)
+        if (
+          kafkaOffsets.beginningOffsets.isEmpty || kafkaOffsets.endOffsets.isEmpty || kafkaOffsets.beginningOffsets.keySet != kafkaOffsets.endOffsets.keySet
+        ) {
```
kevinwallimann: I would log a warning here

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaService.scala
```diff
@@ -22,15 +22,18 @@ import org.springframework.stereotype.Service
 import org.springframework.util.ConcurrentLruCache
 import za.co.absa.hyperdrive.trigger.api.rest.services.KafkaServiceImpl.{BeginningOffsets, EndOffsets, OffsetFunction}
 import za.co.absa.hyperdrive.trigger.configuration.application.GeneralConfig
+import za.co.absa.hyperdrive.trigger.models.BeginningEndOffsets
 
 import java.util.Properties
 import java.util.UUID.randomUUID
 import javax.inject.Inject
 import scala.collection.JavaConverters._
+import scala.util.Try
```
kevinwallimann: unused

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaService.scala
```diff
@@ -50,6 +53,14 @@ class KafkaServiceImpl @Inject() (generalConfig: GeneralConfig) extends KafkaSer
     getOffsets(topic, consumerProperties, EndOffsets)
   }
 
+  def getOffsets(topic: String, consumerProperties: Properties): BeginningEndOffsets = {
```
kevinwallimann: I would rename this to `getBeginningEndOffsets` to be consistent

src/test/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveServiceTest.scala
```diff
@@ -0,0 +1,104 @@
...
+    reset(hyperdriveOffsetService)
+  }
+
+  "getIngestionStatus" should "fail on get workflow failure" in {
+    val id = 1
+    val error = "error"
+    when(workflowRepository.getWorkflow(any())(any())).thenReturn(Future.failed(new Exception(error)))
+
+    val result = the[Exception] thrownBy await(underTest.getIngestionStatus(id))
+    result.getMessage shouldBe error
```
kevinwallimann: Why are you not using an async style here, like in the next test?

ui/src/app/components/workflows/workflows-home/workflows-home.component.html
```diff
@@ -220,6 +220,49 @@ <h3 class="modal-title">Upload workflow</h3>
...
+          <clr-dg-column>Progress</clr-dg-column>
+
+          <clr-dg-placeholder>We couldn't load detail!</clr-dg-placeholder>
+
+          <clr-dg-row *ngFor="let job of ingestionStatus" [clrDgItem]="job">
+            <clr-dg-cell>{{job.jobName}}</clr-dg-cell>
+            <clr-dg-cell>{{job.jobType}}</clr-dg-cell>
+            <clr-dg-cell *ngIf="job?.topic">
+              <b>Topic:</b> {{job?.topic?.topic}} <br/>
+              <div *ngIf="job?.topic?.messagesToIngest >= 0"><b>Messages to ingest:</b> {{job?.topic?.messagesToIngest >= 0 ? job?.topic?.messagesToIngest : 'Inconsistency detected. Please contact support team!'}}</div>
```
kevinwallimann: Can't you omit the ternary operator here, just have `job?.topic?.messagesToIngest`?

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/controllers/HyperdriveController.scala
```diff
@@ -0,0 +1,32 @@
...
+import za.co.absa.hyperdrive.trigger.models._
+
+import java.util.concurrent.CompletableFuture
+import javax.inject.Inject
+import scala.compat.java8.FutureConverters._
+import scala.concurrent.ExecutionContext.Implicits.global
+
+@RestController
+class HyperdriveController @Inject() (hyperdriveService: HyperdriveService) {
+  @GetMapping(path = Array("/hyperdrive/workflows/{id}/ingestionStatus"))
```
kevinwallimann, line 29: The structure of this REST API endpoint is a bit strange, since hyperdrive is not really an entity of us. But our REST API is a mess anyway.

jozefbakus: 👍

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaService.scala
jozefbakus: Done

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaService.scala
jozefbakus: Done

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/CheckpointService.scala
jozefbakus: I don't see a reason why. It is valid flow, if checkpoint folder does not exist.

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveService.scala
jozefbakus: Protected now

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveOffsetService.scala
jozefbakus: Done

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/CheckpointService.scala
jozefbakus: I rewritten this peace multiple times and it is residue of that. Done

src/main/scala/za/co/absa/hyperdrive/trigger/models/IngestionStatus.scala
jozefbakus: Done

src/test/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveServiceTest.scala
jozefbakus: Fixed

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/CheckpointService.scala
jozefbakus: Done

ui/src/app/components/workflows/workflows-home/workflows-home.component.html
jozefbakus: Fixed

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveOffsetService.scala
jozefbakus: Done

kevinwallimann (dismissed):

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveOffsetService.scala
```diff
@@ -52,6 +57,59 @@ class HyperdriveOffsetComparisonServiceImpl @Inject() (sparkConfig: SparkConfig,
...
+    }
+
+    Future(
+      for {
+        kafkaParameters <- kafkaParametersOpt
+        hdfsParameters <- hdfsParametersOpt
+      } yield {
+        val kafkaOffsets = kafkaService.getBeginningEndOffsets(kafkaParameters._1, kafkaParameters._2)
+        if (
+          kafkaOffsets.beginningOffsets.isEmpty || kafkaOffsets.endOffsets.isEmpty || kafkaOffsets.beginningOffsets.keySet != kafkaOffsets.endOffsets.keySet
```
filiphornak: It is more a matter of style. However, for me, it would be much more clear to use match with `unapply` on kafkaOffsets:
```scala
kafkaOffsets match {
  case BeginningEndOffsets(_, start, end) if start.keySet == end.keySet => // do useful things
  case => // Something went wrong
}
```

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveOffsetService.scala
```diff
@@ -52,6 +57,59 @@ class HyperdriveOffsetComparisonServiceImpl @Inject() (sparkConfig: SparkConfig,
...
+        } else {
+          val ugi = userGroupInformationService.loginUserFromKeytab(hdfsParameters.principal, hdfsParameters.keytab)
+          val hdfsOffsetsTry = checkpointService.getLatestCommittedOffset(hdfsParameters)(ugi).map(_.map(_.head._2))
+
+          hdfsOffsetsTry match {
+            case Failure(_) => None
+            case Success(hdfsOffsetsOption) =>
+              val messagesLeft = kafkaOffsets.beginningOffsets.map { case (partition, kafkaBeginningOffset) =>
+                val kafkaEndOffset = kafkaOffsets.endOffsets(partition)
+                val numberOfMessages = hdfsOffsetsOption.flatMap(_.get(partition)) match {
```
filiphornak: Can there occur a situation where `kafkaBeginningOffset > kafkaEndOffset`? Because if not, you might simplify this match pattern to just two cases:
```scala
val numberOfMessages = hdfsOffsetOption.flatMap(_.get(partition)) match {
  case Some(hdfsOffset) if hdfsOffset > kafkaBeginningOffset => kafkaEndOffset - hdfsOffset
  case _                                                     => kafkaEndOffset - kafkaBeginningOffset
}
```
Which I would probably simplify even more:
```scala
val numberOfMessages = kafkaEndOffset - hdfsOffset(partition).map(math.max(_, kafkaBeginningOffset)).getOrElse(kafkaBeginningOffset)

// where hdfsOffset
def hdfsOffset(partition: Long): Option[Long] =
  hdfsOffsetOption.flatMap(_.get(partition))
```

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaService.scala
```diff
@@ -50,6 +52,14 @@ class KafkaServiceImpl @Inject() (generalConfig: GeneralConfig) extends KafkaSer
     getOffsets(topic, consumerProperties, EndOffsets)
   }
 
+  def getBeginningEndOffsets(topic: String, consumerProperties: Properties): BeginningEndOffsets = {
+    BeginningEndOffsets(
+      topic,
+      getOffsets(topic, consumerProperties, BeginningOffsets),
```
filiphornak, line 58: This is a suggestion for future projects because it would require reworking or unification of the entire architecture. I noticed that some services use Futures as output, and some don't even when they are blocking.

```scala
case class Offset(beginning: Long, end: Long)

def getBeginningEndOffset(topic: String, consumerProperties: Properties): Future[Map[Long, Offset]] =
  for {
    kafkaConsumer   <- consumerPool.getConsumer(consumerProperties)
    parts           <- listTopicPartitions(kafkaConsumer, topic)
    futureBeginning  = getBeginningOffsets(kafka, parts)
    futureEnd        = getEndOffsets(kafka, parts)
    begin           <- futureBegin
    end             <- futureEnd
  } yield begin.map { case (part, bOff) => part -> Offset(bOff, end(part)) }  // This might be done in safer manner
```
This might improve performance.

src/main/scala/za/co/absa/hyperdrive/trigger/models/BeginningEndOffsets.scala
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.hyperdrive.trigger.models
+
+case class BeginningEndOffsets(
+  topic: String,
+  beginningOffsets: Map[Int, Long],
```
filiphornak, line 20: You are checking that `beginningOffsets` has the same keys (partition numbers) as `endOffsets`. Could the Kafka partitioning change during that request? Or do you think, maybe, that this service could return `Failure` either as `Future` or `Try` monad?

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/CheckpointService.scala
```diff
@@ -98,6 +102,18 @@ class CheckpointServiceImpl @Inject() (@Lazy hdfsService: HdfsService) extends C
...
   }
 
+  override def getLatestCommittedOffset(
+    params: HdfsParameters
+  )(implicit ugi: UserGroupInformation): Try[Option[TopicPartitionOffsets]] = {
+    getLatestCommitBatchId(params.checkpointLocation).flatMap {
+      case Some(latestCommit) =>
+        val pathToLatestCommit = new Path(s"${params.checkpointLocation}/$offsetsDirName/$latestCommit")
+        getOffsetsFromFile(pathToLatestCommit.toString)
+          .map(_.map(topicPartitionOffsets => topicPartitionOffsets))
```
filiphornak: This seems like a map of map of identity, which is identity. Am I missing something? Did you need this to change widen/narrow types? Because this should be equivalent to calling just `getOffsetFromFile(pathToLatestCommit)`. In my personal opinion, I would use a slightly shorter way:
```scala
getLatestCommitBatchId(params.checkpointLocation).flatMap {
    _.map { latestCommit =>
      getOffsetsFromFile(new Path(params.checkpointLocation, offsetDirName, latestCommit.toString)
    }.getOrElse(Success(None))
  }
```

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveOffsetService.scala
```diff
@@ -52,6 +57,59 @@ class HyperdriveOffsetComparisonServiceImpl @Inject() (sparkConfig: SparkConfig,
...
+      } yield {
+        val kafkaOffsets = kafkaService.getBeginningEndOffsets(kafkaParameters._1, kafkaParameters._2)
+        if (
+          kafkaOffsets.beginningOffsets.isEmpty || kafkaOffsets.endOffsets.isEmpty || kafkaOffsets.beginningOffsets.keySet != kafkaOffsets.endOffsets.keySet
+        ) {
+          logger.warn(s"Inconsistent response from kafka for topic: ${kafkaOffsets.topic}")
+          None
+        } else {
+          val ugi = userGroupInformationService.loginUserFromKeytab(hdfsParameters.principal, hdfsParameters.keytab)
+          val hdfsOffsetsTry = checkpointService.getLatestCommittedOffset(hdfsParameters)(ugi).map(_.map(_.head._2))
```
filiphornak: This is why I like the `cats` library, where you can define `Functor[Try[Option]]` and use that instead:
```scala
val TryOptFunctor = Functor[Try].compose[Option]

TryOptFunctor.map(checkpointService.getLatestCommitedOffset(hdfsParameters)(ugi))(_.head._2)
// or simply
Functor[Try].compose[Option].map(checkpointService.getLatestCommitedOffset(hdfsParameters)(ugi))(_.head._2)
```
Also, I have a question about the usage of the head directly. Isn't there a possibility that the commit file has a version line and metadata line but no offsets or offsets with the invalid or old format? I went through the `HdfsService#parseFileAndClose[R]` method, which returns `Try[R]` and throws an exception when it can not parse such a file. It seems intentional; however, it would be much better to return `Try` or, even better, a `Future`. I know this is outside the scope of this ticket, but it is just something to bear in mind.
```scala
def readFile[R](path: Path, parse: Iterator[String] => Try[R])(implicit ugi: UserGroupInformation): Future[Option[R]] =
  exists(path).flatMap {
    case true => open(path).map(readLines(Codec.UTF8)).flatMap(execAndClose(parse)).flatMap(Future.fromTry)
    case _    => Future.success(None)
  }

def execAndClose[R <: Closeable, O](fn: R => O)(res: => R): Future[O] =
  try {
    Future(fn(res))
  } finally {
    IOUtils.closeQuietly(res)
  }

private def readLines(codec: Codec)(is: InputStream): Iterator[String] =
  Source.fromInputStream(is)(codec).getLines()
```
This draft assumes that open can handle failure properly and release its resources.

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveService.scala
```diff
@@ -0,0 +1,83 @@
...
+      jobTemplateService
+        .resolveJobTemplate(workflow.dagDefinitionJoined)
+        .flatMap(resolvedJobs =>
+          Future.sequence(
+            resolvedJobs.map {
+              case resolvedJob if resolvedJob.jobParameters.jobType == JobTypes.Hyperdrive =>
+                hyperdriveOffsetService.getNumberOfMessagesLeft(resolvedJob.jobParameters).transformWith {
+                  case Failure(exception) =>
+                    logger.error(s"Failed to get number of messages left to ingest for a workflow: $id", exception)
+                    Future(
```
filiphornak: I don't know if this is intentional to use `Future#apply` instead of `Future#successful`. But to me, creating `IngestionStatus` is not an IO-bound operation. However, I just briefly scanned over `ResolvedJobDefinition`, and it seemed that it doesn't lazily evaluate some of its attributes, but again I might be wrong. Here is an article I stumbled upon [https://blog.colinbreck.com/future-successful-blocking-your-future-success/](https://blog.colinbreck.com/future-successful-blocking-your-future-success/)

src/test/scala/za/co/absa/hyperdrive/trigger/api/rest/services/CheckpointServiceTest.scala
```diff
@@ -163,6 +163,88 @@ class CheckpointServiceTest extends FlatSpec with Matchers with BeforeAndAfter w
...
+    when(hdfsService.exists(any())(any())).thenReturn(Success(true))
+    when(hdfsService.listStatus(any(), any())(any())).thenReturn(Success(createOffsetFiles(12)))
+    when(hdfsService.parseFileAndClose[underTest.TopicPartitionOffsets](any(), any())(any()))
+      .thenReturn(Try(Some(offsets)))
+
+    val params = getHdfsParameters
+
+    val result = underTest.getLatestCommittedOffset(params)(ugi)
+
+    result.get.isDefined shouldBe true
```
filiphornak, line 244: Instead, I use the `equals` or `matchesPattern` partial function. Again this is a matter of style.
```scala
result should equal(Success(Some(offsets))) 
// or
result should matchPattern { case Success(Some(_)) => }
```

src/test/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveOffsetServiceTest.scala
```diff
@@ -275,6 +271,140 @@ class HyperdriveOffsetComparisonServiceTest extends AsyncFlatSpec with Matchers
...
+    val config = getSparkConfig
+    val jobParameters = getJobParameters
+    val underTest = new HyperdriveOffsetServiceImpl(config.yarn, checkpointService, ugiService, kafkaService)
+
+    when(ugiService.loginUserFromKeytab(any(), any())).thenReturn(ugi)
+    when(kafkaService.getBeginningEndOffsets(any(), any()))
+      .thenReturn(BeginningEndOffsets("topic", Map.empty, Map.empty))
+
+    val resultFut = underTest.getNumberOfMessagesLeft(jobParameters)
+    resultFut.map { result =>
```
filiphornak, line 284: I'm just curious, can't you simply write `resultFut map (_ shouldBe None)`? However, this is much more readable.

src/test/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaServiceTest.scala
```diff
@@ -73,4 +74,40 @@ class KafkaServiceTest extends FlatSpec with MockitoSugar with Matchers {
...
+    when(mockKafkaConsumer.partitionsFor(any()))
+      .thenReturn(partitions.asJava)
+      .thenReturn(partitions.asJava)
+    when(mockKafkaConsumer.beginningOffsets(eqTo(topicPartitions))).thenReturn(startOffsets)
+    when(mockKafkaConsumer.endOffsets(eqTo(topicPartitions))).thenReturn(endOffsets)
+
+    val result = underTest.getBeginningEndOffsets(topicName, new Properties())
+
+    result shouldBe BeginningEndOffsets(topicName, Map(0 -> 100L, 1 -> 200L), Map(0 -> 200L, 1 -> 400L))
+  }
```
filiphornak, line 104: Do you also want to create tests where the beginning and end offsets return different partitions?

filiphornak (commented):
The code is well-written, at least the scala part. However, there is one inconsistency when reading the last commit file, which may throw an exception even when the method returns the `Try` monad. 

Also, reading offsets from kafka and the latest checkpoint file are not parallelized due to the current implementation of the kafka cache. However, this would require a proper consumer pool implementation because calls on ordinary kafka consumer are not thread-safe or finding alternative how to request offsets from kafka without a consumer.

src/test/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveOffsetServiceTest.scala
jozefbakus: Yes you can, but there are other tests where i check more than one thing where it is not possible. So I am being consistent.

src/main/scala/za/co/absa/hyperdrive/trigger/models/BeginningEndOffsets.scala
jozefbakus: I do, I might happen. I later in the code relay on fact there is same number of partitions, so I rather check it.
If you mean kafka service then no. It is valid response. I don't see a reason why to check it there

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaService.scala
jozefbakus: I agree with you. Please create an issue for this one. With this PR I just follow style that was already introduced

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveOffsetService.scala
jozefbakus: This version is more readable

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveOffsetService.scala
jozefbakus: I prefer this style too, but I think Kevin usually is against. But I rewrote it

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveService.scala
jozefbakus: Cool, changed it to Future.successful

src/test/scala/za/co/absa/hyperdrive/trigger/api/rest/services/CheckpointServiceTest.scala
jozefbakus: Good to know. 
Please create an issue for this, we can do it for all tests and be consistent

src/test/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaServiceTest.scala
jozefbakus: I think I wrote similar test in HyperdriveOffsetServiceTest. Does not make sense to write it here.

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveOffsetService.scala
jozefbakus: In theory, yes it can happen and we handle it. 
Please create an issue for this

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/CheckpointService.scala
jozefbakus: Done

src/test/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaServiceTest.scala
filiphornak: Sorry, I missed that.

src/test/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveOffsetServiceTest.scala
filiphornak: Ok, I understand that

src/main/scala/za/co/absa/hyperdrive/trigger/models/BeginningEndOffsets.scala
filiphornak: Ok, that is reasonable.

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveOffsetService.scala
filiphornak: Ok, I will leave it to you. As I said, this is mainly a matter of style issue.

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/KafkaService.scala
filiphornak: Will do!

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/services/HyperdriveOffsetService.scala
filiphornak: Will do!

src/test/scala/za/co/absa/hyperdrive/trigger/api/rest/services/CheckpointServiceTest.scala
filiphornak: Will do!

filiphornak (approved):
I'm pleased with this explanation and your changes, so I approve this PR. Also, I will create issues for the things we discussed earlier.

PR #805, opened 2023-03-23
#802: Adding some log messages in hyperdrive trigger

### LINKED ISSUE #802: Add more debug log messages to flow activating, deactivating and exection of workflow
## Background
To narrow down the issue location of [https://github.com/AbsaOSS/hyperdrive-trigger/issues/801](https://github.com/AbsaOSS/hyperdrive-trigger/issues/801). Does occur, we have to implement some trace logs in workflow assignment and dag and execution, especially for `TimeSensor`.

## Feature
Add several log messages to the scheduler and sensors that help us investigate the ordering of calls before creating dag, especially those that didn't happen.

filiphornak:
I added some logging messages for the hyperdrive trigger. It will give us insight into what's happening with skipping specific workflows.


src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/utilities/logging/LazyToStr.scala
```diff
@@ -0,0 +1,31 @@
...
+ */
+package za.co.absa.hyperdrive.trigger.scheduler.utilities.logging
+
+/**
+ *  Class deferring computation on toString call.
+ *
+ *  @param value That should be computed only after toString call.
+ *  @tparam A Type of value computation result
+ */
+class LazyToStr[A](value: => A) {
```
filiphornak: This might not be required anymore. The `scala-logging` macro should wrap the call to in
```scala
if (logger.is<Level>Enabled) logger.<level>(message, args)
```

So proper usage should be
```scala
logger.debug("Data = ${heavyComputationOfData()}")
```
However, we need first to check that it behaves this way.

filiphornak: So I checked it out, and it can be replaced directly with `logger.debug(s"... ${value}")`

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/health/DatabaseConnectionHealthIndicator.scala
```diff
@@ -32,7 +32,7 @@ class DatabaseConnectionHealthIndicator @Inject() (val dbProvider: DatabaseProvi
     extends HealthIndicator
     with Repository {
   import api._
-  private val log = LoggerFactory.getLogger(this.getClass)
+  private val log = Logger(this.getClass)
```
jozefbakus, line 35: Why not to do it with LazyLogging trait ?

src/test/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/spark/SparkEmrClusterServiceTest.scala
```diff
@@ -56,7 +56,7 @@ class SparkEmrClusterServiceTest
     with BeforeAndAfter {
   implicit override def executionContext: ExecutionContext = scala.concurrent.ExecutionContext.Implicits.global
 
-  private trait HasUpdateJob {
+  trait HasUpdateJob {
```
jozefbakus: Why not private ?

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/cluster/WorkflowBalancer.scala
```diff
@@ -65,19 +64,20 @@ class WorkflowBalancer @Inject() (
...
+  }
 
   private def getOrCreateInstance()(implicit ec: ExecutionContext) =
     schedulerInstanceId match {
       case Some(id) => Future(id)
       case None =>
         schedulerInstanceService
           .registerNewInstance()
-          .map { id =>
+          .map(wireTap { id =>
```
jozefbakus: I don't like it, I see this idea. But I would leave as it was. It is visible that what is returned and what is a side effect on first scan

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/sensors/Sensors.scala
```diff
@@ -100,25 +100,51 @@ class Sensors @Inject() (
...
     val activeSensors = sensors.keys.toSeq
-    sensorRepository.getInactiveSensors(activeSensors).map(_.foreach(id => stopSensor(id)))
+    logger.trace(
+      "Removing inactive sensors called with active sensors: {}",
+      new LazyToStr(activeSensors.map(id => s"SensorId=$id"))
+    )
+    sensorRepository
+      .getInactiveSensors(activeSensors)
+      .map(
+        wireTap(inactive =>
```
jozefbakus: You can do this in the bellow map without wire tap

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/sensors/Sensors.scala
```diff
@@ -100,25 +100,51 @@ class Sensors @Inject() (
...
+        )
+      )
+      .map(_.foreach(id => stopSensor(id)))
   }
 
   private def stopSensor(id: Long) = {
-    sensors.get(id).foreach(_.close())
+    logger.trace("Stopping sensor (SensorId={})", id)
+    sensors.get(id).foreach { sensor =>
+      logger.debug(
```
jozefbakus: Duplicate, it is logged in sensor.close method

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/eventProcessor/EventProcessor.scala
```diff
@@ -42,27 +42,66 @@ class EventProcessor(
...
+      logger.trace(
+        "Unprocessed events {}",
+        new LazyToStr(newEvents.map(e => s"EventId=${e.id}"))
+      )
       if (newEvents.nonEmpty) {
         dagDefinitionRepository.getJoinedDagDefinition(sensorId).flatMap {
           case Some(joinedDagDefinition) =>
             for {
-              hasInQueueDagInstance <- dagInstanceRepository.hasInQueueDagInstance(joinedDagDefinition.workflowId)
+              hasInQueueDagInstance <- dagInstanceRepository
```
jozefbakus, line 58: You can rewrite this and have logging after each "flatmap"
_ = logger.trace("bla")

src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/health/DatabaseConnectionHealthIndicator.scala
filiphornak: These classes already had a logger provided by slick or another framework named `logger`, but with a different type (`SlickLogger` or something like that), so there was a clash, and it won't compile. It's the same reason why the values are called `log`, `serviceLogger`, etc., in the first place.

src/test/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/spark/SparkEmrClusterServiceTest.scala
filiphornak: I actually don't know, but when I ran it locally, it complained that the mockito framework could not mock `private static` classes. I used the same command as in the GitHub actions script.

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/cluster/WorkflowBalancer.scala
filiphornak: Ok, I will remove it. But, IMO, this looks clearer when you see that there is no change in value, just execution of side-effect.

Akka had something similar in their stream. Even native scala contains `.tap`, or `.tapEach`. And it could be easily implemented on any `Functor` type with `cats`.
```scala
import cats._

implicit class FunctorOps[F[_], A](val fa: F[A]) extends AnyVal {
  def tap[A, U](f: A => U)(implicit func: Functor[F]): F[A] =
    func.map(fa) { elm =>
      f(elm)
      elm
    }
}
```

Because, to me, this looks much cleaner
```scala
schedulerInstanceId match {
  case Some(id) => Future.successful(id)
  case _        => 
    schedulerInstanceService
      .registerNewInstance()
      .tap(id => schedulerInstanceId = Some(id))
      .omComplete {
        case Success(id) =>
          logger.info(s"Successfully assigned new (SchedulerId=${id}) to scheduler")
        case Failure(e) =>
          logger.error("Failed to get new SchedulerId", e)
      }
}
```

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/eventProcessor/EventProcessor.scala
filiphornak: Yeah, I know about it, and I will change it back. However, I don't like it because the `=` is for a map if scala had something for specifying side effects in for comprehension.

I would instead use something like [https://typelevel.org/cats/typeclasses/arrow.html](https://typelevel.org/cats/typeclasses/arrow.html) or a better example with precisely the same thing [https://medium.com/virtuslab/arrows-monads-and-kleisli-part-ii-12ffd4da8bc9](https://medium.com/virtuslab/arrows-monads-and-kleisli-part-ii-12ffd4da8bc9).

To me, writing it that way looks much cleaner, and also, using cats would add several additional structures, like  `IO`, `Functor` (which in turn can be used to simplify all the `.map(_.map(..._.map(f)...))` operations), `Validated`, `Writer`, etc...

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/sensors/Sensors.scala
filiphornak: Yeah, ok, I will remove it with `LazyToStr` (because with the change to `scala-logging`, it's also not required anymore)

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/sensors/Sensors.scala
filiphornak: Thank you! I'll get rid of it.

src/test/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/spark/SparkEmrClusterServiceTest.scala
filiphornak: So I reverted it and will test it in the pipeline

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/Executors.scala
```diff
@@ -127,16 +130,20 @@ class Executors @Inject() (
           }
         }
         fut.onComplete {
-          case Success(_) => logger.debug(s"Executing job. Job instance = $jobInstance")
+          case Success(_) => logger.debug(s"Executing job. (JobId={})", jobInstance)
```
jozefbakus: JobInstance instead of JobId

jozefbakus: The same bellow

src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/Executors.scala
```diff
@@ -127,16 +130,20 @@ class Executors @Inject() (
...
-            logger.error(s"Executing job failed. Job instance id = $jobInstance.", exception)
+            logger.error(s"Executing job failed. (JobId=$jobInstance).", exception)
         }
         fut
     }
 
   private def updateJob(jobInstance: JobInstance): Future[Unit] = {
     logger.info(
-      s"Job updated. ID = ${jobInstance.id} STATUS = ${jobInstance.jobStatus} EXECUTOR_ID = ${jobInstance.executorJobId}"
+      "(JobId={}). Job updated. ID = {} STATUS = {} EXECUTOR_ID = {}",
```
jozefbakus: Unnecessary job id twice

jozefbakus (approved):

PR #816, opened 2023-11-22
S3a support for ingestion status

jozefbakus:
S3a support for ingestion status

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -113,7 +113,8 @@
         <hibernate.validator.version>6.2.0.Final</hibernate.validator.version>
         <aws.java.sdk.bom.version>1.12.29</aws.java.sdk.bom.version>
         <absa.commons.version>1.0.3</absa.commons.version>
-
+        <hadoop-aws.version>2.7.3.2.6.1.0-129</hadoop-aws.version>
+        <aws-java-sdk-bundle.version>1.11.375</aws-java-sdk-bundle.version>
         <!-- Deployment -->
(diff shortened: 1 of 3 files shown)


src/main/scala/za/co/absa/hyperdrive/trigger/api/rest/WebMvcConfig.scala
```diff
@@ -17,11 +16,21 @@
...
+  WebMvcConfigurer
+}
 
 @Configuration
 class WebMvcConfig extends WebMvcConfigurer {
   override def addViewControllers(registry: ViewControllerRegistry): Unit = {
     registry.addViewController("/").setViewName("forward:/index.html")
   }
+
+  override def configureAsyncSupport(configurer: AsyncSupportConfigurer): Unit = {
```
kevinwallimann, line 31: What is this needed for?

pom.xml
```diff
@@ -442,6 +444,16 @@
             <version>${absa.commons.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-aws</artifactId>
+            <version>${hadoop-aws.version}</version>
+        </dependency>
+        <dependency>
```
kevinwallimann, line 452: Why is this needed explicitly? Is it not already a dependency of hadoop-aws?

kevinwallimann (approved):

PR #827, opened 2024-01-11
Feature/819 store error messages

### LINKED ISSUE #819: Store and send error message of failed applications
## Background
Currently, an e-mail notification is sent to subscribed users. If an application failed, the URL to the Application's page on the resource manager is added to the notification.

However, the application's page may be retained only for a short while on the cluster (e.g. 1-2 days), after which the error message cannot be retrieved anymore. Moreover, if there are multiple failed applications, it's cumbersome to always click the link. Furthermore, no analysis is currently possible on error patterns that frequently occur.

Therefore, in this feature, the error message should be retrieved from the resource manager when an application has failed and stored in the database. When a notification for a failed application is sent, this error message should be added to the e-mail message.

The size of an error message is expected to be around 10-30 KB.

## Feature
- [x] Add a new optional text column in the `job_instance` table, `error_message`
- [x] Also add it to the `archive_job_instance` table
- [x] Update the `archive_dag_instances_chunk` procedure to include the new column
- [ ] Update the `JobInstanceTable` and the `JobInstance` case class
- [ ] In the `SparkExecutor`, add logic to retrieve the error message in case of a job failure and save it in the JobInstance
- [ ] In the `NotificationSender` add logic to print the error message along the RM's application url in case of a job failure

## Example
The error message for an application can be retrieved e.g. like this:
`curl http://{{rm-http-address}}:8088/ws/v1/cluster/apps/{{application-id}} | jq -r '.app.diagnostics'`

kevinwallimann:
Closes #819 

How it was tested locally:

application.properties changed properties:
```
notification.enabled=true
notification.sender.address=<redacted>
spring.mail.host=<redacted>
spring.mail.port=<redacted>
spring.mail.properties[mail.smtp.connectiontimeout]=5000
spring.mail.properties[mail.smtp.timeout]=5000
spring.mail.properties[mail.smtp.writetimeout]=5000

sparkYarnSink.hadoopResourceManagerUrlBase=http://<redacted.:8088

db.driver=org.postgresql.Driver
db.url=jdbc:postgresql://localhost:5432/hyperdriver
db.user=hyperdriver
db.password=hyperdriver
```

- Create dummy workflow, click run
- Job Instance should be stuck in Submitting status.
- Create a suitable notification rule (E.g. receive for all projects and prefixes). Don’t forget to activate it
- Copy a real executor job id from the test system and copy it to your job instance. Also set the job status to ‘Running’. E.g.
`UPDATE job_instance SET (job_status, executor_job_id) VALUES (‘Running’,<real-executor-job-id>) WHERE id = <your-local-job-instance-id>`
- Refresh job instance table (directly in DB), see stack trace filled in the diagnostics column of your job instance.
- Wait a couple of minutes, receive e-mail notification with stacktrace.

I did it both for failed and successful runs, and also for a failed run whose application page was already empty

After these tests were finished, finally I executed
```
CALL archive_dag_instances(
        i_to_ts => now()::timestamp,
        i_max_records => 200000,
        i_chunk_size => 10000
     );
```

### DIFF
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/notifications/NotificationSender.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/notifications/NotificationSender.scala
@@ -122,23 +122,24 @@ class NotificationSenderImpl(
         val applicationUrl = s"${yarnBaseUrl.stripSuffix("/")}/cluster/app/$appId"
         messageMap += ("Failed application" -> applicationUrl)
       })
-    val diagnosticsOpt = failedJob.flatMap(_.diagnostics)
-    diagnosticsOpt.map { diagnostics =>
-      causedByPattern.findFirstMatchIn(diagnostics) match {
-        case Some(m) => messageMap += ("Caused by" -> m.group(1))
-        case None => // do nothing
-      }
-    }
+
     messageMap += ("Notification rule ID" -> notificationRule.id.toString)
 
-    val diagnosticsDetail = diagnosticsOpt.map { diagnostics =>
-      s"Job diagnostics:\n$diagnostics\n\n"
+    val diagnosticsOpt = failedJob.flatMap(_.diagnostics)
+    val causes = diagnosticsOpt.map { diagnostics =>
+      causedByPattern.findAllMatchIn(diagnostics).map(_.group(1)).toSeq.reduce(_ + "\n" + _)
+    }
+      .map("Causes:\n" + _ + "\n\n")
+      .getOrElse("")
+
+    val stackTrace = diagnosticsOpt.map { diagnostics =>
+      s"Stack trace:\n$diagnostics\n\n"
     }.getOrElse("")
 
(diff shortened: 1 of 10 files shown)


src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/notifications/NotificationSender.scala
```diff
@@ -114,15 +115,40 @@ class NotificationSenderImpl(
...
+      messageMap += ("Failed application" -> applicationUrl)
+    })
+
     messageMap += ("Notification rule ID" -> notificationRule.id.toString)
-    val message = messageMap.map { case (key, value) => s"$key: $value" }.reduce(_ + "\n" + _) + "\n\n" + footer
+
+    val diagnosticsOpt = failedJob.flatMap(_.diagnostics)
+    val causes = diagnosticsOpt
+      .map { diagnostics =>
+        causedByPattern
```
jozefbakus, line 131: I think this one is overkill. I would just add whole text within diagnosticsOpt.

jozefbakus (approved):

PR #828, opened 2024-01-22
#818: Regularly execute archive_dag_instances function

### LINKED ISSUE #818: Regularly execute archive_dag_instances function
## Background
In #721, the `archive_dag_instances` SQL function was added. It moves the oldest dag instances (and related objects) to an archive table, so that querying the `dag_instance` doesn't slow down over time. By default, the last 6 months are kept in the `dag_instance` table and are queriable by end users.

This query can take a long time (minutes - 10s of minutes), depending on the number of objects to be removed. However, executing it more frequently should reduce the execution time of a single query to seconds and not impact the database.

Currently, the `archive_dag_instances` function needs to be invoked manually. This is obviously not ideal. Therefore, a scheduled job should invoke the function in a regular interval (e.g. every Sunday).

## Feature
- Create a cron-job (e.g. with the Quartz library) that calls the `archive_dag_instances` function.
- The cron-job should take retention days as a parameter and use it for the second parameter of the `archive_dag_instances` function, which is `i_to_ts`. `i_to_ts` should be set to `now() - '${retentionDays} days'::interval`
- The output of the `archive_dag_instances` function should be captured and sent to the configurable notification recipients.
- The total execution time of the query should be reported as well. It doesn't have to be exact, it can also include the roundtrip time to the database.
- New application parameters to add
  - `housekeeping.archiveDags.schedule` (default `0 0 4 ? * SUN *`)
  - `housekeeping.archiveDags.retentionDays` (default 180)
  - `housekeeping.archiveDags.notificationRecipients` (default [])

- Only one of the scheduler instances should execute the function! This could be achieved using an optimistic locking technique with a auxiliary database table or other techniques. While skipping an archival is not catastrophic, the archive function should not be called concurrently.

jozefbakus:
Closes: #818 

Run by executing: `./cleanUpRuns.sh  "email_for_notifications" "retention_days"  "env"`
It requires command-line utilities: mail(xmail) and psql

### DIFF
--- a/src/main/resources/db_scripts/housekeeping/cleanUpRuns.sh
+++ b/src/main/resources/db_scripts/housekeeping/cleanUpRuns.sh
@@ -1,4 +1,18 @@
 #! /bin/bash
+#
+# Copyright 2018 ABSA Group Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
 
 RECIPIENTS=$1
 RETENTION_DAYS=$2
--- a/src/main/resources/db_scripts/housekeeping/cleanUpRuns.sh
+++ b/src/main/resources/db_scripts/housekeeping/cleanUpRuns.sh
@@ -17,6 +17,7 @@
 RECIPIENTS=$1
 RETENTION_DAYS=$2
 ENV=$3
+CONFIG_FILE_LOCATION=${4:-"../../application.properties"}
 
 if [[ -n "$RECIPIENTS" && -n "$RETENTION_DAYS" && -n "$ENV" ]]; then
   echo "Starting runs clean up with retention days: $RETENTION_DAYS, env: $ENV and recipients: $RECIPIENTS"
@@ -25,11 +26,9 @@ else
   exit 1
 fi
 
-config_file="../../application.properties"
-
-db_url=$(grep '^db.url=' "$config_file" | awk -F'jdbc:postgresql://' '{print $2}' | awk -F'\\?' '{print $1}')
-db_user=$(grep '^db.user=' "$config_file" | awk -F'=' '{print $2}')
-db_password=$(grep '^db.password=' "$config_file" | awk -F'=' '{print $2}')
+db_url=$(grep '^db.url=' "$CONFIG_FILE_LOCATION" | awk -F'jdbc:postgresql://' '{print $2}' | awk -F'\\?' '{print $1}')
+db_user=$(grep '^db.user=' "$CONFIG_FILE_LOCATION" | awk -F'=' '{print $2}')
+db_password=$(grep '^db.password=' "$CONFIG_FILE_LOCATION" | awk -F'=' '{print $2}')
 connection_url="postgresql://$db_user:$db_password@$db_url"
 
 obtain_lock_response=$(psql -X $connection_url -c "UPDATE housekeepinglock SET locked = 'true', started_at = now() WHERE locked = false AND started_at is null;" 2>&1)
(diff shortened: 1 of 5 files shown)


src/main/resources/db_scripts/housekeeping/cleanUpRuns.sh
```diff
@@ -0,0 +1,78 @@
...
+ENV=$3
+
+if [[ -n "$RECIPIENTS" && -n "$RETENTION_DAYS" && -n "$ENV" ]]; then
+  echo "Starting runs clean up with retention days: $RETENTION_DAYS, env: $ENV and recipients: $RECIPIENTS"
+else
+  echo "Incorrect input. Exiting the program."
+  exit 1
+fi
+
+config_file="../../application.properties"
```
kevinwallimann: Maybe use an absolute path here? Or where do you expect this shell-script to be located?

src/main/resources/db_scripts/housekeeping/cleanUpRuns.sh
```diff
@@ -0,0 +1,78 @@
...
+if [[ $obtain_lock_response == "UPDATE 1" ]]; then
+  echo "Exclusive lock for running runs clean up job obtained."
+
+  archive_dag_instances_response=$(psql -X $connection_url -c "CALL archive_dag_instances(i_to_ts => (now() - '$RETENTION_DAYS days'::interval)::timestamp without time zone);" 2>&1)
+  release_lock_response=$(psql -X $connection_url -c "UPDATE housekeepinglock SET locked = 'false', started_at = null;" 2>&1)
+
+  message=""
+  subject=""
+  if [[ $archive_dag_instances_response == *"ERROR"* ]]; then
+    subject="Hyperdrive Notifications - $ENV - Runs clean up succeeded"
```
kevinwallimann: should be failed, not succeeded. Vice versa below

src/main/resources/db_scripts/housekeeping/cleanUpRuns.sh
```diff
@@ -0,0 +1,78 @@
...
+else
+  obtain_lock_time_response=$(psql -X $connection_url -c "SELECT started_at FROM housekeepinglock;" 2>&1)
+
+  timestamp_string=$(echo "$obtain_lock_time_response" | grep -oE '[0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}:[0-9]{2}')
+  timestamp_unix=$(date -jf "%Y-%m-%d %H:%M:%S" "$timestamp_string" +%s)
+  current_unix=$(date +%s)
+  time_difference=$((current_unix - timestamp_unix))
+  days_difference=$((time_difference / 86400))
+
+  if [ "$days_difference" -gt "29" ]; then
```
kevinwallimann: I would reduce the threshold to e.g. 2 days

src/main/resources/db_scripts/housekeeping/cleanUpRuns.sh
```diff
@@ -0,0 +1,78 @@
...
+    message="Runs clean up has been running for more than 30 days."
+    echo -e "$message"
+    echo -e "$message" | mailx -s "Hyperdrive Notifications - $ENV - Runs clean up locked" -r "noreply@absa.co.za" "$RECIPIENTS"
+  else
+    echo "Runs clean up is running on different machine."
```
kevinwallimann: It could also be that it was running on the same machine before it was restarted or that the other machine couldn't release the lock, so I would rephrase this. I would probably print out the time difference

src/main/resources/db_scripts/housekeeping/cleanUpRuns.sh
jozefbakus: I did it this way now: 
CONFIG_FILE_LOCATION=${4:-"../../application.properties"}
So if user inputs path, that path will be used.
If no path is on input script will assume that is running from original location.

src/main/resources/db_scripts/housekeeping/cleanUpRuns.sh
jozefbakus: Done

src/main/resources/db_scripts/housekeeping/cleanUpRuns.sh
jozefbakus: Done

src/main/resources/db_scripts/housekeeping/cleanUpRuns.sh
jozefbakus: Done

kevinwallimann (approved):

kevinwallimann:
Required for deployment:
- cronjob
- installation of mail and psql if needed

PR #841, opened 2024-03-06
Add saveDiagnostics application property

### LINKED ISSUE #840: Add application property to switch on storage of error messages
## Background
In #819, a new feature was added that stores error messages of failed runs in the database. Since the database could grow significantly because of this feature, it should only be enabled after a regular cleanup of the job instances table has been scheduled (see #828 and https://github.com/absa-group/ansible_projects-hyperdrive_automated_deployment/issues/17) 

## Feature
- Add a new boolean application property `sparkYarnSink.saveDiagnostics` should be added, default value: `false`
- If the property is set to true, save the diagnostics, otherwise don't

kevinwallimann:
Closes #840 

Tested locally with `sparkYarnSink.saveDiagnostics` set to true, false, and without setting the property at all.

**New Application property**
- `sparkYarnSink.saveDiagnostics`. Valid values: `true`, `false`. Default value: `false`

### DIFF
--- a/src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/spark/SparkExecutor.scala
+++ b/src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/spark/SparkExecutor.scala
@@ -74,9 +74,9 @@ object SparkExecutor {
     app: App
   )(implicit sparkConfig: SparkConfig): JobInstance = {
     val diagnostics = app.diagnostics match {
-      case "" => None
+      case ""                                => None
       case _ if !sparkConfig.saveDiagnostics => None
-      case _  => Some(app.diagnostics)
+      case _                                 => Some(app.diagnostics)
     }
 
     jobInstance.copy(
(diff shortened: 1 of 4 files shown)


src/main/scala/za/co/absa/hyperdrive/trigger/scheduler/executors/spark/SparkExecutor.scala
```diff
@@ -72,10 +72,11 @@ object SparkExecutor {
   private def getUpdatedJobInstance(
     jobInstance: JobInstance,
     app: App
-  ): JobInstance = {
+  )(implicit sparkConfig: SparkConfig): JobInstance = {
     val diagnostics = app.diagnostics match {
-      case "" => None
-      case _  => Some(app.diagnostics)
+      case ""                                => None
+      case _ if !sparkConfig.saveDiagnostics => None
```
jozefbakus: I would formulate that positively.
And "default" case left with None - always fallback to None.

`
val diagnostics = app.diagnostics match {
      case "" => None
      case _ if sparkConfig.saveDiagnostics  => Some(app.diagnostics)
      case _  => None
    }
`

jozefbakus (approved):

PR #850, opened 2025-07-31
Fixed hadoop, hadoop-aws and aws-java-sdk lib versions

jozefbakus:
Fixed hadoop, hadoop-aws and aws-java-sdk lib versions

When hadoop version 2.7.3.2.6.1.0-129 is used, it has to be used with aws-java-sdk 1.10.6 and hadoop-aws 2.7.3.2.6.1.0-129


kevinwallimann (approved):
lgtm

